PR #13, opened 2020-08-14
Money type

cscederborg:


centrallix-lib/include/datatypes.h
```diff
@@ -56,8 +56,11 @@ typedef struct _SV
 /** Money structure **/
 typedef struct _MN
     {
-    int                 WholePart;      /* integer part = floor($amount) */
-    unsigned short      FractionPart;   /* fract part = $amount - floor($amount) */
+    /* Commented out old structure */
+    //int                 WholePart;      /* integer part = floor($amount) */
+    //unsigned short      FractionPart;   /* fract part = $amount - floor($amount) */
+    
+    long long       MoneyValue;     /* Fixed-point representation in 1/10000 of a dollar */
```
sheesania, line 63: I think Greg wanted this to just be "Value"

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -490,18 +490,29 @@ obj_internal_FormatMoney(pMoneyType m, char* str, char* format, int length)
...
-	    print_whole = m->WholePart + 1;
-	    print_fract = 10000 - m->FractionPart;
+	    { //This handled the unsigned representation of FractionPart
+	    print_whole = m->MoneyValue/10000 + 1;
+	    print_fract = 10000 - m->MoneyValue%10000;
 	    }
-	orig_print_whole = m->WholePart;
-	if (print_whole < 0) print_whole = -print_whole;
+	orig_print_whole = m->MoneyValue/10000;
+	if (print_whole < 0) print_whole = -print_whole;*/
```
sheesania, line 515: don't forget to remove commented-out stuff once you're done testing

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -520,22 +531,22 @@ obj_internal_FormatMoney(pMoneyType m, char* str, char* format, int length)
                     *(str++) = '$';
                     break;
    
-    		case ' ':
-		case '*':
+    		    case ' ':
```
sheesania, line 534: revert these tab changes at some point so they don't clutter up the diffs

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -802,10 +813,12 @@ objDataToInteger(int data_type, void* data_ptr, char* format)
 
 	    case DATA_T_MONEY: 
 	        m = (pMoneyType)data_ptr;
-                if (m->FractionPart==0 || m->WholePart>=0)
-		    v = m->WholePart;
+	        v = m->MoneyValue/10000;
```
sheesania, line 816: check that m->MoneyValue/10000 isn't larger than INT_MAX

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -840,7 +853,7 @@ objDataToDouble(int data_type, void* data_ptr)
 	    {
 	    case DATA_T_INTEGER: v = *(int*)data_ptr; break;
 	    case DATA_T_STRING: v = strtod((char*)data_ptr, NULL); break;
-	    case DATA_T_MONEY: m = (pMoneyType)data_ptr; v = m->WholePart + (m->FractionPart/10000.0); break;
+	    case DATA_T_MONEY: m = (pMoneyType)data_ptr; v = m->MoneyValue/10000 + ((m->MoneyValue%10000)/10000.0); break;
```
sheesania, line 856: If you can do double arithmetic, you can just divide MoneyValue by 10000 without doing any other arithmetic

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1401,7 +1414,7 @@ objDataToMoney(int data_type, void* data_ptr, pMoneyType m)
     char* endptr2;
     double dbl;
     int is_neg = 0;
-    unsigned long intval;
+    long long intval;
```
sheesania, line 1417: double check intval isn't used elsewhere in the function with the assumption that it's signed/possibly fewer bits??

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1442,26 +1455,33 @@ objDataToMoney(int data_type, void* data_ptr, pMoneyType m)
 		    ptr++;
 		    }
 		intval = 0;
-		m->FractionPart = 0;
-		m->WholePart = 0;
-		intval = strtoul(ptr, &endptr, 10);
-		if (intval > 0x7FFFFFFFUL)
+		m->MoneyValue = 0;
+		intval = strtoll(ptr, &endptr, 10);
```
sheesania, line 1459: Will the string ever be parsed as a negative integer?

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1442,26 +1455,33 @@ objDataToMoney(int data_type, void* data_ptr, pMoneyType m)
...
+		if (intval > 0x7FFFFFFFFFFFFFFFLL)
```
sheesania, line 1460: This check shouldn't be necessary anymore since you're putting a signed long long into a signed long long (no change in signing). But you should still somehow handle strings that would overflow

sheesania, line 1460: Yay comments!

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1442,26 +1455,33 @@ objDataToMoney(int data_type, void* data_ptr, pMoneyType m)
...
-		    intval = strtoul(endptr+1, &endptr2, 10);
+		    intval = strtoll(endptr+1, &endptr2, 10);
 		    scale = endptr2 - (endptr+1);
 		    if (scale != strspn(endptr+1, "0123456789"))
 			return -1;
 		    while(scale < 4) { scale++; intval *= 10; }
 		    while(scale > 4) { scale--; intval /= 10; }
-		    m->FractionPart = intval;
+		    if (is_neg)
+                {
```
sheesania, line 1477: maybe double check whether you're using the same spaces/tabs as the surrounding code?

sheesania: and whether CLion is doing autoformatting

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1470,30 +1490,29 @@ objDataToMoney(int data_type, void* data_ptr, pMoneyType m)
...
+		    }*/
 		break;
 	    
 	    case DATA_T_DOUBLE:
 	        dbl = *(double*)data_ptr + 0.000001;
-	        m->WholePart = floor(dbl);
-		m->FractionPart = (dbl - m->WholePart)*10000;
+	        long long WholePart = floor(dbl);
+		    int FractionPart = (dbl - WholePart)*10000;
+		    m->MoneyValue = (WholePart*10000) + FractionPart;
```
sheesania, line 1507: You should be able to just multiply the double by 10000

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1470,30 +1490,29 @@ objDataToMoney(int data_type, void* data_ptr, pMoneyType m)
...
 		break;
 	
 	    case DATA_T_INTEGER:
-	        m->WholePart = *(int*)data_ptr;
-		m->FractionPart = 0;
+	        m->MoneyValue = *(long long*)data_ptr;
```
sheesania, line 1511: You should multiply the int by 10000 so it represents the whole part, as in the original version

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1617,9 +1636,9 @@ objDataCompare(int data_type_1, void* data_ptr_1, int data_type_2, void* data_pt
 		    case DATA_T_MONEY:
 		        objDataToMoney(DATA_T_STRING, data_ptr_1, &m_v);
 			m = (pMoneyType)data_ptr_2;
-			if (m_v.WholePart > m->WholePart) cmp_value = 1;
-			else if (m_v.WholePart < m->WholePart) cmp_value = -1;
-			else cmp_value = m_v.FractionPart - m->FractionPart;
+			if (m_v.MoneyValue/10000 > m->MoneyValue/10000) cmp_value = 1;
+			else if (m_v.MoneyValue/10000 < m->MoneyValue/10000) cmp_value = -1;
+			else cmp_value = m_v.MoneyValue%10000 - m->MoneyValue%10000;
```
sheesania, line 1641: you should be able to just compare MoneyValue without separately comparing the whole/fraction sections

sheesania: and then you can just set cmp_value to 0 in the final else

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1673,7 +1692,7 @@ objDataCompare(int data_type_1, void* data_ptr_1, int data_type_2, void* data_pt
 
 		    case DATA_T_MONEY:
 			m = (pMoneyType)data_ptr_2;
-		        dblval = m->WholePart + (m->FractionPart/10000.0);
+		        dblval = m->MoneyValue/10000 + ((m->MoneyValue%10000)/10000.0);
```
sheesania, line 1695: You can probably just divide by 10000 to get a double here too

sheesania: 10000.0 to be precise :)

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1821,9 +1840,9 @@ objDataCompare(int data_type_1, void* data_ptr_1, int data_type_2, void* data_pt
 		switch(data_type_2)
 		    {
 		    case DATA_T_MONEY: 
-			if (m->WholePart > ((pMoneyType)data_ptr_2)->WholePart) cmp_value = 1;
-			else if (m->WholePart < ((pMoneyType)data_ptr_2)->WholePart) cmp_value = -1;
-			else cmp_value = m->FractionPart - ((pMoneyType)data_ptr_2)->FractionPart;
+			if (m->MoneyValue/10000 > ((pMoneyType)data_ptr_2)->MoneyValue/10000) cmp_value = 1;
+			else if (m->MoneyValue/10000 < ((pMoneyType)data_ptr_2)->MoneyValue/10000) cmp_value = -1;
+			else cmp_value = m->MoneyValue%10000 - ((pMoneyType)data_ptr_2)->MoneyValue%10000;
```
sheesania, line 1845: same as above, you don't need to do divide or modulus, and just set cmp_value to 0 in the last else

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1470,30 +1479,21 @@ objDataToMoney(int data_type, void* data_ptr, pMoneyType m)
...
-		    m->WholePart--;
-		    m->FractionPart = 10000 - m->FractionPart;
-		    }
 		break;
 	    
 	    case DATA_T_DOUBLE:
-	        dbl = *(double*)data_ptr + 0.000001;
-	        m->WholePart = floor(dbl);
-		m->FractionPart = (dbl - m->WholePart)*10000;
+            m->Value = *(double*)data_ptr * 10000.0;
```
sheesania, line 1488: A couple things:
- Do you know why the original implementation adds 0.000001 to the double? Is that something you should still do?
- Are you sure that m->Value is going to end up as a proper long long type this way, or does it need to be multiplied by 10000 (without the decimal place) or cast?

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1574,9 +1574,9 @@ objDataCompare(int data_type_1, void* data_ptr_1, int data_type_2, void* data_pt
 
 		    case DATA_T_MONEY:
 		        m = (pMoneyType)data_ptr_2;
-			if (m->WholePart > intval) cmp_value = -1;
-			else if (m->WholePart < intval) cmp_value = 1;
-			else cmp_value = m->FractionPart?-1:0;
+			if (m->Value/10000 > intval) cmp_value = -1;
+			else if (m->Value/10000 < intval) cmp_value = 1;
+			else cmp_value = m->Value%10000?-1:0;
```
sheesania, line 1579: It would probably be simpler to just compare (intval * 10000) to m->Value, returning -1, 1, or 0 as appropriate.

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1769,9 +1769,9 @@ objDataCompare(int data_type_1, void* data_ptr_1, int data_type_2, void* data_pt
 			    }
```
sheesania, line 1769: Hmm, it looks like the old implementation expects to be given 2 integers, so it can compare the first integer to WholePart and the second integer to FractionPart. With the new MoneyType representation, it might make more sense to only expect one integer to compare to, but I'm not sure - I think we should ask Greg about it when he gets back. Maybe leave a comment about this so we remember to ask about that?

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -2184,11 +2177,10 @@ obj_internal_BuildBinaryItem(char** item, int* itemlen, pExpression exp, pParamO
 		break;
 
 	    case DATA_T_MONEY:
-		/** XOR 0x80000000 to convert to Offset Zero form. **/
-		((unsigned int*)tmp_buf)[0] = htonl(exp->Types.Money.WholePart ^ 0x80000000);
-		((unsigned short*)tmp_buf)[2] = htons(exp->Types.Money.FractionPart);
+		/** XOR 0x8000000000000000 to convert to Offset Zero form. **/
+		((long long*)tmp_buf)[0] = bswap_64(exp->Types.Money.Value ^ 0x8000000000000000);
```
sheesania, line 2181: Can you add a more descriptive comment about what this is doing?

centrallix/tests/test_moneytype_00.c
```diff
@@ -0,0 +1,30 @@
+#include "obj.h"
+#include <assert.h>
+
+long long
+test(char** name)
+{
+    *name = "moneytype_00 - obj_internal_FormatMoney";
```
sheesania, line 7: Since this is technically testing objFormatMoneyTmp, not obj_internal_FormatMoney, could you change the name to reflect that?

centrallix/tests/test_moneytype_01.c
```diff
@@ -0,0 +1,44 @@
...
+#include <stdio.h>
+//#include "xstring.h"
+//#include "mtsession.h"
+//#include "mtask.h"
+
+long long
+test(char** name)
+{
+    *name = "moneytype_01 - objDataToInteger";
+    int testValue;
```
sheesania, line 12: Could you give this variable a more descriptive name (like `testInt` or something else indicating that it's the integer output), or just remove it (e.g. `assert(objDataToInteger(7,&test,NULL) == 0)`)?

centrallix/tests/test_moneytype_01.c
```diff
@@ -0,0 +1,44 @@
...
+    /** Positive Case **/
+    test.Value = 10000;
+    testValue = objDataToInteger(7,&test,NULL);
+    assert(testValue == 1);
+
+    /** Negative Case **/
+    test.Value = -10000;
+    testValue = objDataToInteger(7,&test,NULL);
+    assert(testValue == -1);
+
```
sheesania, line 28: Can you add a case for a MoneyType with cents/fractional value?

centrallix/tests/test_moneytype_01.c
```diff
@@ -0,0 +1,44 @@
...
+    /** Overflow Case **/
+    //Manual Testing showed this section to work properly on overflow
+    //Currently, I do not know how to print the error messages, so this part
```
sheesania, line 31: Are you blocked on this or do you still have some ideas for how you might be able to write this test? If you're blocked I can do some poking around myself

centrallix/tests/test_moneytype_02.c
```diff
@@ -0,0 +1,31 @@
+#include "obj.h"
+#include <assert.h>
+
+long long
+test(char** name)
+{
+    *name = "moneytype_02 - objDataToDouble";
+    double testValue;
```
sheesania, line 8: as above, I think it would be more clear if this was something like `testDouble`, or removed entirely

centrallix/tests/test_moneytype_04.c
```diff
@@ -0,0 +1,40 @@
+#include "obj.h"
+#include <assert.h>
+#include <stdio.h>
+
+long long
+test(char** name)
+{
+    *name = "moneytype_04 - objDataCompare";
+    MoneyType test = {70500};
+
```
sheesania, line 10: Can you add more cases for less than/greater than/equals for all these types? You can also break the tests out into multiple .c files if you feel like it's getting long.

centrallix/tests/test_moneytype_05.c
```diff
@@ -0,0 +1,25 @@
...
+    char* data_ptr = "Seven And 05/100 ";
+    char* returnStr = objDataToWords(7,&test);
+    assert(strcmp(data_ptr, returnStr) == 0);
+    
+    /** Negative Case **/
+    test.Value = -70500;
+    data_ptr = "Negative Seven And 05/100 ";
+    returnStr = objDataToWords(7,&test);
+    assert(strcmp(data_ptr, returnStr) == 0);
+
```
sheesania, line 23: Can you add a case for fractional cents? E.g. if you can't express the cents in hundredths

centrallix/tests/test_moneytype_03.c
```diff
@@ -0,0 +1,36 @@
...
+long long
+test(char** name)
+{
+    *name = "moneytype_03 - objDataToMoney";
+    MoneyType test = {0};
+
+    /** String Case **/
+    char data_ptr[] = "$4.50";
+    assert(objDataToMoney(2,data_ptr,&test) == 0);
+    assert(test.Value == 45000);
```
sheesania, line 14: can you add a case for a string representing a negative money value?

centrallix/tests/test_moneytype_03.c
```diff
@@ -0,0 +1,36 @@
...
+
+    /** Double Case **/
+    double testDouble = 5.5;
+    assert(objDataToMoney(3,&testDouble,&test) == 0);
+    assert(test.Value == 55000);
```
sheesania, line 19: Would it be possible for a double to have more significant digits in the fractional part than can be stored in a MoneyType? If so could you add a case for that?

A case for a negative double would also be great

centrallix/tests/test_moneytype_03.c
```diff
@@ -0,0 +1,36 @@
...
+    assert(test.Value == 60000);
+    
+    /** Money Case **/
+    MoneyType testMoney = {900000};
+    assert(objDataToMoney(7,&testMoney,&test) == 0);
+    assert(test.Value == 900000);
+
+    /** Overflow Case (intval > LL max) **/
+    char overflow_ptr[] = "$10000000000000000000.50";
+    assert(objDataToMoney(2,overflow_ptr,&test) == -1);
```
sheesania, line 33: It would probably be clearest if this test case was next to the other string case(s), so it's more obvious that's a special case for strings specifically

centrallix/objectsystem/obj_datatypes.c
sheesania: Scratch this, I just realized this is for comparing an _IntVec_ type to a MoneyType, not an Integer type 🤪

centrallix/tests/test_moneytype_00.c
cscederborg: The changes were to obj_internal_FormatMoney, and objFormatMoneyTmp is basically a wrapper that allocates a buffer and calls obj_internal_FormatMoney.  The point of the test is to see if ob_internal_FormatMoney is working properly, even though it calls objFormatMoneyTmp.  I'll add a comment explaining that.

centrallix/tests/test_moneytype_04.c
```diff
@@ -0,0 +1,40 @@
...
+    MoneyType testMoney = {70500};
+    assert(objDataCompare(7,&testMoney,7,&test) == 0);
+
+    /** Compare DateTime with Money Case **/
+    DateTime testDate = {0};
+    assert(objDataCompare(4,&testDate,7,&test) == -2);
+
+    /** Compare IntV with Money Case **/
+    IntVec testIV = {0};
+    assert(objDataCompare(5,&testIV,7,&test) == -2);
```
sheesania, line 33: Can you add cases for where there are two ints in the IntVec and the comparison goes through?

centrallix/tests/test_moneytype_00.c
sheesania: One unit testing best practice I forgot to mention to you is that you generally want to test the public interface - who cares how it's implemented under the hood so long as the functions that will be called elsewhere give you the right results. (If you can change your implementation without breaking your clients, your unit tests also shouldn't break!) So in this case, it doesn't really matter that objFormatMoneyTmp works under the hood by calling obj_internal_FormatMoney, or that you had to edit that function specifically. All an external caller cares is that objFormatMoneyTmp spits out what it's supposed to, so that should be the focus of testing. 

So you wrote these tests right by calling objFormatMoneyTmp, but I think it would be best for the name to also reflect that. Does that make sense?

centrallix/expression/exp_evaluate.c
```diff
@@ -320,7 +309,11 @@ expEvalDivide(pExpression tree, pParamObjects objlist)
 			    mssError(1,"EXP","Attempted divide by zero");
 			    return -1;
 			    }
-			mv = ((long long)(i0->Types.Money.WholePart)) * 10000 + i0->Types.Money.FractionPart;
+			md = i0->Types.Money.Value / 10000.0;
+			md = md / i1->Types.Double;
+			tree->Types.Money.Value = (long long)(md * 10000);
+			//Old Definition
+			/*mv = ((long long)(i0->Types.Money.WholePart)) * 10000 + i0->Types.Money.FractionPart;
```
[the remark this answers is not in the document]
sheesania: Looking at this again...I think my initial comment was wrong and your old code DID handle remainders (since it's dividing i0->Types.Money.Value / 10000.0 into a double type, so it should end up with a double with decimal places, etc). It seems like the adding/subtracting 0.5 might actually be related to the old representation's handling of negative values? (See also the adding 10000 to mv and subtracting one from WholePart on line 333 in the old version.) So you probably don't need to add/subtract 0.5?? Does that make sense to you?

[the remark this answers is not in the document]
cscederborg: Yeah it makes sense.  I'll comment it out for now and see how it goes once I get to expression testing.

centrallix/utility/obfuscate.c
```diff
@@ -1211,6 +1212,7 @@ obfObfuscateData(pObjData srcval, pObjData dstval, int data_type, char* attrname
 		    }
 		//m.WholePart = floor(dv/100.0);
 		//m.FractionPart = (dv - m.WholePart*100) * 100;
+		m.Value = dv * 100;
```
sheesania, line 1215: The original code is pretty gnarly - you got it right as far as I can tell, but just make sure you run tests on the old + new versions to make sure you get the same results!

centrallix/osdrivers/objdrv_fp.c
```diff
@@ -1227,7 +1227,7 @@ fp_internal_ParseColumn(pFpColInf column, pObjData pod, char* data, char* row_da
 	    case DATA_T_INTEGER:
 		if (fp_internal_MappedCopy(ibuf, sizeof(ibuf), column, row_data) < 0) return -1;
 		pod->Integer = strtoi(ibuf, NULL, 10);
-		break;
+		break
```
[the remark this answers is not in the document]
cscederborg: Must have been an accident, I'm switching it back.

centrallix/utility/json_util.c
```diff
@@ -236,11 +236,13 @@ jutilGetMoneyObject(struct json_object* jobj, pMoneyType m)
 		    {
 		    has_whole = 1;
 		    //m->WholePart = json_object_get_int(iter.val);
+		    m->Value += json_object_get_int(iter.val) * 10000;
```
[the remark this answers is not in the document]
cscederborg: Documentation seems to suggest that.  I could throw in a line setting Value to zero for safety.

centrallix-lib/src/net_raw_api.c
```diff
@@ -192,7 +192,7 @@ rawSendParam(pNetManagedConn conn, char param_type, pObjData value, int cmd_id)
 		break;
 
 	    case NET_PARAM_MONEY:
-	        memcpy(buf+1, value->Money, 6);
+	        memcpy(buf+1, value->Money, 8);
 		len = 7;
```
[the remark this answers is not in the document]
cscederborg: It is setting up the buffer which will be written with fdWrite later on line 210.  I changed from 6 to 8 because old representation was int+unsigned short and new is long long.  I also now changed len to 9, because len is size+1 in all the cases; this change will show in my next push.

centrallix/netdrivers/net_http_rest.c
```diff
@@ -114,10 +114,10 @@ nht_i_RestWriteAttrValue(pNhtConn conn, pObject obj, char* attrname, int data_ty
 		break;
 
 	    case DATA_T_MONEY:
-		/*Carl nht_i_QPrintfConn(conn, 0, "{ \"wholepart\":%INT, \"fractionpart\":%INT }",
-			od.Money->WholePart,
-			od.Money->FractionPart
-			);*/
+		nht_i_QPrintfConn(conn, 0, "{ \"wholepart\":%INT, \"fractionpart\":%INT }",
```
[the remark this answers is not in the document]
cscederborg: Looking at the documentation, it doesn't seem to be able to handle larger than normal int values.  I think my best bet would be simply checking beforehand and failing the function if it is too big.  There is not a % tag that handles long longs.

centrallix/tests/test_moneytype_fp_internal_ParseColumn_00.c
```diff
@@ -7,23 +7,34 @@
 long long
 test(char** name)
 {
-    *name = "moneytype_06 - obj_internal_BuildBinaryItem";
+    *name = "moneytype_00 - fp_internal_ParseColumn";
```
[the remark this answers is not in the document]
cscederborg: Yes, correct.

centrallix/tests/test_moneytype_objDataCompare_00.c
```diff
@@ -6,7 +6,7 @@
 long long
 test(char** name)
 {
-    *name = "moneytype_04 - objDataCompare";
+    *name = "moneytype_00 - objDataCompare";
```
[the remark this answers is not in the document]
cscederborg: Some of the other .to and .cmp files, even if they were for one function, still had _00 at the end, so I followed suit.

centrallix/tests/test_moneytype_obj_internal_BuildBinaryItem_00.c
```diff
@@ -0,0 +1,63 @@
...
+    ObjData moneyData;
+    MoneyType unionFiller = {70000};
+    moneyData.Money = &unionFiller;
+    pObjData moneyDataPtr = &moneyData;
+    
+    //Utilizing expression.h functions to initialize expression and paramobjects
+    pExpression testExp = expAllocExpression();
+    pParamObjects testParamObjects = expCreateParamList();
+    
+    //Buffer must be at least 12 bytes
```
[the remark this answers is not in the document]
cscederborg: No it was a comment in his code so I relayed the message here.  I figure it has to do with the range of things that can be passed into it.  Potentially something else is a big enough struct that requires 12 bytes.  Looking at the code, there is no itemlen of 12, only up to 9, so it could be a safety thing too.

centrallix/tests/test_moneytype_objDataCompare_00.c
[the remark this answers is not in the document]
sheesania: sounds good then

centrallix/tests/test_moneytype_obj_internal_BuildBinaryItem_00.c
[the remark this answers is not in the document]
sheesania: Ah okay. Can you add something to your comment here then like "the documentation says" or "a comment in the function says" or something like that? Just so it's clear where that idea is coming from.

centrallix/netdrivers/net_http_rest.c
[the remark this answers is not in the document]
gbeeley: No, the %INT will not accept an int64.  Varargs in C just throws untyped / untagged stuff on the stack and expects the implementation to know what it's doing.  int and long long will use different amounts of stack, which means you'll end up being off by 4 bytes for subsequent parameters.  We'll need to add a 64-bit integer % type for qprintf.

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1470,30 +1479,23 @@ objDataToMoney(int data_type, void* data_ptr, pMoneyType m)
...
-		    m->FractionPart = 10000 - m->FractionPart;
-		    }
 		break;
 	    
 	    case DATA_T_DOUBLE:
-	        dbl = *(double*)data_ptr + 0.000001;
-	        m->WholePart = floor(dbl);
-		m->FractionPart = (dbl - m->WholePart)*10000;
+            dbl = *(double*)data_ptr * 10000.0;
+            dbl = round(dbl);
```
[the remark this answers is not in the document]
gbeeley: And... mysql also **does** round:

```
mysql> select convert(1.005, decimal(10,2));
+-------------------------------+
| convert(1.005, decimal(10,2)) |
+-------------------------------+
|                          1.01 |
+-------------------------------+
1 row in set (0.00 sec)

mysql> select convert(1.004, decimal(10,2));
+-------------------------------+
| convert(1.004, decimal(10,2)) |
+-------------------------------+
|                          1.00 |
+-------------------------------+
1 row in set (0.00 sec)

```

centrallix/osdrivers/objdrv_fp.c
```diff
@@ -1243,18 +1244,15 @@ fp_internal_ParseColumn(pFpColInf column, pObjData pod, char* data, char* row_da
...
+		for (i=0; i<column->DecimalOffset; i++) decimalOffsetValue /= 10.0;
 		pod->Money = (pMoneyType)data;
-		pod->Money->WholePart = v/f;
-		v = (v/f)*f;
-		if (column->DecimalOffset <= 4)
-		    for(i=column->DecimalOffset;i<4;i++) v *= 10;
-		else
-		    for(i=4;i<column->DecimalOffset;i++) v /= 10;
-		pod->Money->FractionPart = v;
+		pod->Money->Value = v*decimalOffsetValue;
```
gbeeley, line 1255: The challenge here is that this cannot be straightforwardly tested due to no real way to do column type autodetection - all columns are tagged as DATA_T_STRING.  So even with test data we can't trigger this segment of code without handcrafting the data structures.

I would recommend using pow() instead of the repeated division, as pow() may have a better way of managing error as opposed to the cumulative error from repeated division on a floating point type.  Also, when assigning to ->Value, add 0.1 to the product of v*decimalOffsetValue, also dealing with potential floating point error (like 10.0 becoming 9.999999999999 and then truncating to 9 for the long long assignment).

centrallix/osdrivers/objdrv_sybase.c
```diff
@@ -807,17 +808,20 @@ sybd_internal_GetCxValue(void* ptr, int ut, pObjData val, int datatype)
...
 		if (minus)
-		    {
-		    val->Money->WholePart = -val->Money->WholePart;
-		    if (val->Money->FractionPart > 0)
-			{
-			val->Money->WholePart--;
-			val->Money->FractionPart = 10000 - val->Money->FractionPart;
-			}
-		    }
+		    val->Money->Value = -val->Money->Value;
```
gbeeley, line 813: The original objdrv_sybase.c code here was written in a world where there was no int64 / long long, and the 8-byte value at **ptr** at line 786 in the original code (memcpy(&lsl, ptr+4, 4)) is an int64 (long long), most significant little endian int32 first and least significant little endian int32 second.  So it's mixed-endian.  But once the mixed-endianness is corrected, the value is correct as a long long and is in 10000ths of a currency unit, facilitating a direct assignment to ->Value.

Thus the original code had to divide an int64 by 10000 without int64 support.  That is why it is complex, and does long division one 16-bit word at a time.  The code can be vastly simplified now. :)

centrallix/tests/test_moneytype_obj_internal_BuildBinaryItem_00.c
```diff
@@ -0,0 +1,63 @@
+#include "obj.h"
+#include <assert.h>
+#include <stdio.h>
+#include "expression.h"
+#include "cxlib/xstring.h"
+
+long long
+test(char** name)
+{
+    *name = "moneytype_00 - obj_internal_BuildBinaryItem";
```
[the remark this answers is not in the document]
gbeeley: The binary items/images are supposed to be treated as opaque binary data, used for keying and sorting.  So, to test, I'd recommend encoding some data values (and sets of data values), and ensuring pairwise memcmp() on the resulting data gives the right answer.  see obj_internal_SortCompare() for how to do this, respecting different lengths of the binary data.

Testing the exact data returned may break future refactoring of these functions. :)  The relationships between the returned data values are what is crucial.

centrallix/utility/json_util.c
```diff
@@ -228,19 +228,22 @@ jutilGetMoneyObject(struct json_object* jobj, pMoneyType m)
...
 	json_object_object_foreachC(jobj, iter)
 	    {
 	    if (json_object_is_type(iter.val, json_type_int))
 		{
 		if (!strcmp(iter.key, "wholepart"))
 		    {
 		    has_whole = 1;
-		    m->WholePart = json_object_get_int(iter.val);
+		    //m->WholePart = json_object_get_int(iter.val);
+		    m->Value += json_object_get_int(iter.val) * 10000;
```
gbeeley, line 240: Should be 10000LL, as json_object_get_int returns an int32.

gbeeley (changes requested):
Awesome work, Carl, we're blessed by your help this fall.

centrallix/objectsystem/obj_datatypes.c
gbeeley: This works, but you need to add 0.1 (if positive, or subtract 0.1 if negative) to the value **before** assignment to the long long in the event that `*(double*)data_ptr * 10000.0` ends up something like 9999.999999999999 instead of 10000.0 for example.  Floating point arithmetic is imprecise in a way that can cause this sort of round-off error, and conversion back to an integer value needs a fudge factor included to prevent unexpected truncation.

centrallix/expression/exp_evaluate.c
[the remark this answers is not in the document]
cscederborg: So, I can omit the commented out section, and I also should go and simplify the code where it uses is_negative in this function?

centrallix/netdrivers/net_http_rest.c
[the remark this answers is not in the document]
cscederborg: I am assuming the change will not be as simple as following these steps.  For example, in qpfPrintf_va_internal(), there are a series of switch statements, at least one of which (line 950) will need a case to handle the %LL specifier.  The comment above that function may not be enough for me to understand what I am adding.

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1899,24 +1901,17 @@ objDataToWords(int data_type, void* data_ptr)
...
+            integer_part = -m->Value/10000;
+            fraction_part = -m->Value%10000;
 		xsConcatenate(&tmpbuf, "Negative ", -1);
 		}
 	    else
 	        {
-		integer_part = m->WholePart;
-		fraction_part = m->FractionPart;
+		integer_part = m->Value/10000;
+		fraction_part = m->Value%10000;
```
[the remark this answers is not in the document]
cscederborg: As in, the fraction part was potentially larger than m->Value%10000?

centrallix/expression/exp_evaluate.c
```diff
@@ -649,25 +628,16 @@ expEvalMinus(pExpression tree, pParamObjects objlist)
 		    {
 		    case DATA_T_INTEGER:
 		        tree->DataType = DATA_T_MONEY;
-			tree->Types.Money.WholePart = i0->Types.Money.WholePart - i1->Integer;
-			tree->Types.Money.FractionPart = i0->Types.Money.FractionPart;
+		        /** Since Value is in 1/10000 of a dollar, integer must be multiplied to scale **/
+			tree->Types.Money.Value = i0->Types.Money.Value - (i1->Integer * 10000);
```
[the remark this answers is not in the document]
cscederborg: I have added the long long constant, but I am leaving this comment unresolved to keep in mind the test item.  I can resolve it if it clutters the pull request.

centrallix/netdrivers/net_http_rest.c
```diff
@@ -115,8 +115,8 @@ nht_i_RestWriteAttrValue(pNhtConn conn, pObject obj, char* attrname, int data_ty
 
 	    case DATA_T_MONEY:
 		nht_i_QPrintfConn(conn, 0, "{ \"wholepart\":%INT, \"fractionpart\":%INT }",
-			od.Money->WholePart,
-			od.Money->FractionPart
+			od.Money->Value/10000,
+			od.Money->Value%10000
```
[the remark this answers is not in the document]
cscederborg: Once the %LL specifier is defined, this issue will resolve, correct?  Because after that, I should change the arguments to "'moneyValue: %LL', od.Money->Value".  Unless we want to keep the wholepart/fractionpart split for nht_i_QPrintfConn.

centrallix/objectsystem/obj_datatypes.c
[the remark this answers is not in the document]
gbeeley: The old representation defined the whole part as floor(value) and the fraction part as value - floor(value). This is not the same as using % to get the remainder when you're working with negative values. With the original representation, -$0.0001 would be represented as { -1, 9999 }, not { 0, -1 }, i.e., the fraction part always moves the value in the positive direction, rather than away from zero.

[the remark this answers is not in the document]
cscederborg: Right.  Upon review of the function in the old and the new, doesn't  m->Value%10000 return the proper value for FractionPart, even though it is not equivalent to the old representation?  The old representation set "fraction_part = 10000 - m->FractionPart;", and 10000-m->FractionPart is the same as m->Value%10000.  I think this was a function where I did not change the logic but tried to format the same inputs.

Also a note, objDataToWords() truncates past normal cent value for the string representation, so if that is being used to display how much money someone has, there might be discrepancies between the surface representation and what is actually stored, since we have decided upon rounding for the logic

centrallix/osdrivers/objdrv_fp.c
sheesania: So am I correctly understanding you that this code would never actually run in a real environment - i.e. unless you call this internal function directly with handcrafted data - since columns will always just be tagged as DATA_T_STRING? If so, could you clarify why all this code still exists for other data types and why it should be updated for the new MoneyType, if it's not being used? Are there plans to implement column type autodetection sometime, or is this code never going to be used and should maybe just be removed...?

gbeeley: Correct, it would never run in a production environment.  If y'all would prefer to remove the non-DATA_T_STRING cases, or at least #if 00 them out, you may go ahead (though I'm not concerned about its presence) _after_ we merge this branch into master, so the work done here is in the permanent history and we can get it back & test it at a later date if needed.

sheesania: Okay, thanks for clarifying. It would be great to talk a bit more sometime about the design rationale behind that approach, but that can wait :)

centrallix/osdrivers/objdrv_sybase.c
cscederborg: Is the mixed-endianness corrected in the existing code during the long division section?  Or do I need to write new code for that fix before assigning to Value?

centrallix/netdrivers/net_http_rest.c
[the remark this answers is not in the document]
gbeeley: Yeah, the comment "Get source" isn't too verbose....  I can take a closer look when you have this in a commit.  Basically you have a long long that gets converted to a string (using snprintf is acceptable for this, as is done with QPF_SPEC_T_DBL), to which strval (string value) points, and cplen (copy length) is set to the length of the pointed-to string (not including the null terminator).

BTW, __builtin_expect() is a compiler directive that tells the compiler how to optimize for CPU branch prediction when emitting the assembly code.  The Linux kernel uses LIKELY() and UNLIKELY() macros to wrap this - which we probably should do too sometime... :)

centrallix/netdrivers/net_http_rest.c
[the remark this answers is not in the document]
gbeeley: I would like for us to migrate the JSON REST API to using a single numeric value, but we have to make some design decisions first with that.  For now, the JSON data still should have the whole/fraction split, and it should keep the old representation standard of the fraction part always moving the value in the positive direction.

centrallix/objectsystem/obj_datatypes.c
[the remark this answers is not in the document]
gbeeley: Good point, Carl, the above two lines of code only run when the value is positive.

centrallix/osdrivers/objdrv_fp.c
gbeeley: The main thing to me is to ensure the modified code in question doesn't get squashed out in the merge, and to try to separate the money type work from the removal of this unreachable section of code.

centrallix/osdrivers/objdrv_sybase.c
gbeeley: The existing code deals with it via the msl and lsl values - which are LE 32 bit values.  In your case, when you copy the buffer value into the long long, recognize that you'll need to swap the two 32 bit halves.

I realize this can't be tested without a SAP Sybase ASE server, and we don't have one set up on our servers for testing right now.  I'll do the testing of this on a client's staging instance.  SAP Sybase ASE for Linux is free, so it'd be good for us to get one set up at some point.

centrallix-lib/tests/test_qprintf_64.c
```diff
@@ -0,0 +1,47 @@
...
+        buf[41] = 0xff;
+        buf[40] = '\0';
+        buf[3] = '\n';
+        buf[2] = '\0';
+        buf[1] = 0xff;
+        buf[0] = '\0';
+        qpfPrintf(NULL, buf+4, 36, "Here is the long long: %LL...", 12345ll);
+        qpfPrintf(NULL, buf+4, 36, "Here is the long long: %LL...", 12345ll);
+        qpfPrintf(NULL, buf+4, 36, "Here is the long long: %LL...", 12345ll);
+        rval = qpfPrintf(NULL, buf+4, 36, "Here is the long long: %LL...", 12345ll);
```
gbeeley, line 31: I'd like to see a test with a long long value larger than 0x7FFFFFFF.

centrallix/expression/exp_evaluate.c
```diff
@@ -320,23 +293,11 @@ expEvalDivide(pExpression tree, pParamObjects objlist)
...
-			mv = md;
-			tree->Types.Money.WholePart = mv/10000;
-			mv = mv % 10000;
-			if (mv < 0)
-			    {
-			    mv += 10000;
-			    tree->Types.Money.WholePart -= 1;
-			    }
-			tree->Types.Money.FractionPart = mv;
+			tree->Types.Money.Value = (long long)(i0->Types.Money.Value/i1->Types.Double);
```
gbeeley, line 296: Old logic rounds here, so should new logic.  Use llround() instead of typecast.

centrallix/netdrivers/net_http_rest.c
```diff
@@ -114,10 +115,28 @@ nht_i_RestWriteAttrValue(pNhtConn conn, pObject obj, char* attrname, int data_ty
...
+	        /** For the time being, 64-bit representation will mimic whole and fraction part handling of negatives
+	         ** with regards to JSON data.  It will use the new %LL specifier, but the JSON returned will
+	         ** contain an inverted fraction part that always counts in the positive direction.
+	         **/
+	        
+	        /** If there is a negative with a non-zero fraction part **/
+	        if (od.Money->Value < 0 && od.Money->Value%10000 != 0)
+                {
+	                whole = (od.Money->Value + 10000)/10000ll;
+	                fraction = 10000ll - od.Money->Value%10000;
```
gbeeley, line 127: This math doesn't look right.  For example, for value -9999 (-$0.9999), whole should be -1 and fraction should be 1.  For value -10001 (-$1.0001), whole should be -2 and fraction should be 9999.

centrallix/osdrivers/objdrv_sybase.c
```diff
@@ -795,29 +797,16 @@ sybd_internal_GetCxValue(void* ptr, int ut, pObjData val, int datatype)
...
-		val->Money->WholePart = n;
-		val->Money->FractionPart = msl;
+		/** Both msl and lsl are little endian, so I want lsl to come before msl.
+		** I do this by shifting lsl left in a long long (ll_lsl) by 32 bits, and then OR it with msl
+		**/
+		ll_msl = msl;
+		ll_lsl = lsl;
+		ll_lsl = ll_lsl << 32;
+		ll_lsl = ll_lsl | ll_msl;
+		val->Money->Value = ll_lsll;
```
gbeeley, line 807: Undefined symbol ll_lsll here (I think you meant ll_lsl), also msl is the most significant 32-bit value, and should be left shifted instead of lsl.  I think you can just do val->Money->Value = (((long long)msl)<<32) | (long long)lsl;

centrallix/tests/test_moneytype_objDataToMoney_00.c
```diff
@@ -0,0 +1,55 @@
...
+    int testInt = 6;
+    assert(objDataToMoney(1,&testInt,&test) == 0);
+    assert(test.Value == 60000);
+    
+    /** Money Case **/
+    MoneyType testMoney = {900000};
+    assert(objDataToMoney(7,&testMoney,&test) == 0);
+    assert(test.Value == 900000);
+
+
```
gbeeley, line 52: I'd like some cases (both string and double, both positive and negative) with values that fit in the new Money representation but which overflow 32 bits.

centrallix/utility/obfuscate.c
```diff
@@ -1200,18 +1201,24 @@ obfObfuscateData(pObjData srcval, pObjData dstval, int data_type, char* attrname
...
-		else
-		    {
-		    if (obf_internal_GetBits(hash, &bitcnt, 1)) iv = (iv + 1)*10;
-		    if (obf_internal_GetBits(hash, &bitcnt, 1)) iv = iv/10 + 1;
-		    dv = obf_internal_ObfuscateInteger(hash, hash_novalue, &bitcnt, iv);
-		    }
-		m.WholePart = floor(dv/100.0);
-		m.FractionPart = (dv - m.WholePart*100) * 100;
-		dstval->Money = &m;
+	        if (srcval->Money->Value < INT_MAX)
```
gbeeley, line 1204: Check for negatives, also the max value to check here is: Value < INT_MAX / 100.

gbeeley (changes requested):
Thanks Carl!  We're almost there. :)

centrallix/utility/obfuscate.c
cscederborg: I will place a check for Value to be above INT_MIN/100 as well.  Is that what you mean by check for negatives?

gbeeley: Check to ensure the resulting value can fit in an int32, both at the positive and negative end of the range.

cscederborg: I added the check at the beginning to see if srcval->Money->Value is greater than INT_MIN.  By resulting value, do you mean the value placed into m.Value at the end after the obfuscation process, dv?

gbeeley: No, the intermediate int32 value that gets passed to the obfuscation routine.

cscederborg: Since iv is set to srcval->Money->Value * 100 and I check if srcval->Money->Value is within int max and min /100, isn't the intermediate value already screened to be within an int32?

gbeeley: Yes, the main point is to avoid overflow going into iv, not to constrain what the final m.Value is.

centrallix/netdrivers/net_http_rest.c
```diff
@@ -76,6 +76,7 @@ nht_i_RestWriteAttrValue(pNhtConn conn, pObject obj, char* attrname, int data_ty
     {
     ObjData od;
     int rval;
+        long long whole, fraction;
```
sheesania, line 79: spacing?

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -5,6 +5,7 @@
 #include <time.h>
 #include <stdlib.h>
 #include <math.h>
+#include <byteswap.h>
```
sheesania, line 8: confirm availability

centrallix/osdrivers/objdrv_dbl.c
```diff
@@ -1654,13 +1654,15 @@ dbl_internal_ParseColumn(pDblColInf column, pObjData pod, char* data, char* row_
 		f = 1;
 		for(i=0;i<column->DecimalOffset;i++) f *= 10;
 		pod->Money = (pMoneyType)data;
-		pod->Money->WholePart = v/f;
+		pod->Money->Value = (v/f) * 10000;
+		//pod->Money->WholePart = v/f;
```
sheesania, line 1658: remove comments

centrallix/osdrivers/objdrv_sybase.c
```diff
@@ -775,9 +776,10 @@ sybd_internal_GetCxValue(void* ptr, int ut, pObjData val, int datatype)
 		{
 		/** smallmoney, 4-byte **/
 		memcpy(&i, ptr, 4);
-		val->Money->WholePart = i/10000;
-		if (i < 0 && (i%10000) != 0) val->Money->WholePart--;
-		val->Money->FractionPart = i - (val->Money->WholePart*10000);
+		val->Money->Value = i;
+		//val->Money->WholePart = i/10000;
```
sheesania, line 780: remove comments

centrallix/tests/test_moneytype_obfuscate_00.c
```diff
@@ -0,0 +1,38 @@
...
+#include <assert.h>
+#include <stdio.h>
+#include "expression.h"
+#include "cxlib/xstring.h"
+#include "obfuscate.h"
+
+long long
+test(char** name)
+{
+    *name = "moneytype_00 - obfuscate";
```
sheesania, line 11: Need to confirm that this test is done / working

centrallix/tests/test_moneytype_objDataToInteger_00.c
```diff
@@ -0,0 +1,44 @@
...
+    /** Negative Case **/
+    test.Value = -10000;
+    assert(objDataToInteger(7,&test,NULL) == -1);
+    
+    /** Fractional Case **/
+    test.Value = 15000;
+    assert(objDataToInteger(7,&test,NULL) == 1);
+
+    /** Overflow Case **/
+    //Manual Testing showed this section to work properly on overflow
```
sheesania, line 30: Fix

centrallix/tests/test_moneytype_objDataToInteger_00.c
```diff
@@ -0,0 +1,44 @@
+#include "obj.h"
+#include <assert.h>
+#include <stdio.h>
+//#include "xstring.h"
```
sheesania, line 4: remove comments

centrallix/tests/test_moneytype_objFormatMoneyTmp_00.c
```diff
@@ -0,0 +1,34 @@
+#include "obj.h"
+#include <assert.h>
+
+long long
+test(char** name)
+{
+    /** objFormatMoneyTmp is a wrapper that calls obj_internal_FormatMoney
```
sheesania, line 7: Rework this comment

centrallix/tests/test_sample_money.to
```diff
@@ -0,0 +1,2 @@
+##NAME Money Test 2
```
sheesania, line 1: are these just tests/samples or should they stay in?

centrallix/expression/exp_compiler.c
```diff
@@ -152,8 +152,7 @@ exp_internal_CompileExpression_r(pLxSession lxs, int level, pParamObjects objlis
 			if (t == MLX_TOK_INTEGER)
 			    {
 			    i = mlxIntVal(lxs);
-			    etmp->Types.Money.WholePart = i;
-			    etmp->Types.Money.FractionPart = 0;
+			    etmp->Types.Money.Value = (i * 10000ll);
```
sheesania, line 155: Is there a test that covers this?

centrallix/expression/exp_evaluate.c
```diff
@@ -227,7 +226,7 @@ expEvalDivide(pExpression tree, pParamObjects objlist)
 	/** Check for divide by zero **/
 	if ((i1->DataType == DATA_T_INTEGER && i1->Integer == 0) ||
 	    (i1->DataType == DATA_T_DOUBLE && i1->Types.Double == 0.0) ||
-	    (i1->DataType == DATA_T_MONEY && i1->Types.Money.WholePart == 0 && i1->Types.Money.FractionPart == 0))
+	    (i1->DataType == DATA_T_MONEY && i1->Types.Money.Value == 0))
```
sheesania, line 229: Are there tests covering all these functions?

centrallix/expression/exp_functions.c
```diff
@@ -266,24 +266,11 @@ int exp_fn_abs(pExpression tree, pParamObjects objlist, pExpression i0, pExpress
...
-		    tree->Types.Money.FractionPart = i0->Types.Money.FractionPart;
-		    }
-		else
-		    {
-		    if (i0->Types.Money.FractionPart != 0)
-		        {
-			tree->Types.Money.WholePart = -(i0->Types.Money.WholePart + 1);
-			tree->Types.Money.FractionPart = 10000 - i0->Types.Money.FractionPart;
-			}
+	        if (i0->Types.Money.Value >= 0)
```
sheesania, line 269: Are there tests covering all these functions?

centrallix/expression/exp_main.c
```diff
@@ -523,7 +523,7 @@ exp_internal_DumpExpression_r(pExpression this, int level)
 	    case EXPR_N_INTEGER: printf("INTEGER = %d", this->Integer); break;
 	    case EXPR_N_STRING: printf("STRING = <%s>", this->String); break;
 	    case EXPR_N_DOUBLE: printf("DOUBLE = %.2f", this->Types.Double); break;
-	    case EXPR_N_MONEY: printf("MONEY = %d.%4.4d", this->Types.Money.WholePart, this->Types.Money.FractionPart); break;
+	    case EXPR_N_MONEY: printf("MONEY = %4.4lld", this->Types.Money.Value); break;
```
sheesania, line 526: Are there tests covering this stuff?

centrallix/netdrivers/net_http_rest.c
sheesania, line 79: Are there tests for these changes?

centrallix/osdrivers/objdrv_dbl.c
```diff
@@ -1654,13 +1654,15 @@ dbl_internal_ParseColumn(pDblColInf column, pObjData pod, char* data, char* row_
 		f = 1;
 		for(i=0;i<column->DecimalOffset;i++) f *= 10;
 		pod->Money = (pMoneyType)data;
-		pod->Money->WholePart = v/f;
+		pod->Money->Value = (v/f) * 10000;
```
sheesania, line 1657: tests?

centrallix/osdrivers/objdrv_sybase.c
```diff
@@ -686,6 +686,7 @@ sybd_internal_GetCxValue(void* ptr, int ut, pObjData val, int datatype)
     {
     int i,minus,n;
     unsigned int msl,lsl,divtmp;
+    long long ll_msl, ll_lsl;
```
sheesania, line 689: tests?

centrallix/utility/json_util.c
```diff
@@ -228,19 +228,20 @@ jutilGetMoneyObject(struct json_object* jobj, pMoneyType m)
 
 	/** Search the object's properties **/
 	memset(m, 0, sizeof(MoneyType));
+	m->Value = 0;
```
sheesania, line 231: tests?

sheesania (changes requested):
Hello future self

centrallix/expression/exp_compiler.c
sheesania: Doesn't seem like there is, but it's not worth the effort to try to build one.

centrallix/expression/exp_evaluate.c
```diff
@@ -320,23 +293,11 @@ expEvalDivide(pExpression tree, pParamObjects objlist)
...
-			mv = md;
-			tree->Types.Money.WholePart = mv/10000;
-			mv = mv % 10000;
-			if (mv < 0)
-			    {
-			    mv += 10000;
-			    tree->Types.Money.WholePart -= 1;
-			    }
-			tree->Types.Money.FractionPart = mv;
+			tree->Types.Money.Value = llround(i0->Types.Money.Value/i1->Types.Double);
```
sheesania, line 296: Test rounding?

centrallix/expression/exp_functions.c
```diff
@@ -1936,22 +1916,15 @@ int exp_fn_truncate(pExpression tree, pParamObjects objlist, pExpression i0, pEx
 	    break;
 
 	case DATA_T_MONEY:
-	    mt = ((long long)(i0->Types.Money.WholePart)) * 10000 + i0->Types.Money.FractionPart;
+	    mt = i0->Types.Money.Value;
```
sheesania, line 1919: truncate needs to be tested

centrallix/expression/exp_functions.c
```diff
@@ -3098,8 +3069,7 @@ int exp_fn_max(pExpression tree, pParamObjects objlist, pExpression i0, pExpress
 	    tree->String[0] = '\0';
 	    tree->Integer = 0;
 	    tree->Types.Double = 0;
-	    tree->Types.Money.FractionPart = 0;
-	    tree->Types.Money.WholePart = 0;
+	    tree->Types.Money.Value = 0;
```
sheesania, line 3072: Should max be tested?

centrallix/expression/exp_functions.c
```diff
@@ -3160,8 +3130,7 @@ int exp_fn_min(pExpression tree, pParamObjects objlist, pExpression i0, pExpress
 	    tree->Alloc = 0;
 	    tree->Integer = 0;
 	    tree->Types.Double = 0;
-	    tree->Types.Money.FractionPart = 0;
-	    tree->Types.Money.WholePart = 0;
+	    tree->Types.Money.Value = 0;
```
sheesania, line 3133: Should min be tested?

centrallix-lib/src/qprintf.c
```diff
@@ -58,47 +58,48 @@
 #define QPF_SPEC_T_DBL		(4)
 #define QPF_SPEC_T_NSTR		(5)
 #define QPF_SPEC_T_CHR		(6)
-#define QPF_SPEC_T_ENDSRC	(6)
+#define QPF_SPEC_T_ENDSRC	(7)
+#define QPF_SPEC_T_LL       (7)
```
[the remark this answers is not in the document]
gbeeley: If you'd like to, yes that's fine.  The moneytype PR would need to depend on the %LL PR.

centrallix/expression/exp_functions.c
gbeeley: Yes, max() and min() should be tested, last summer the test suite guys focused on the string functions.

centrallix/expression/exp_main.c
sheesania: This function needs to be tested, probably using (expDumpExpression)[https://sourcegraph.com/github.com/LightSys/centrallix@1b27fd61a93915507c25d7170016112732b4c953/-/blob/centrallix/expression/exp_main.c#L578:1]

centrallix/expression/exp_main.c
```diff
@@ -863,7 +863,7 @@ expCompareExpressionValues(pExpression exp1, pExpression exp2)
 	    return 0;
 	if (!(exp1->Flags & EXPR_F_NULL) && exp1->DataType == DATA_T_DOUBLE && exp1->Types.Double != exp2->Types.Double)
 	    return 0;
-	if (!(exp1->Flags & EXPR_F_NULL) && exp1->DataType == DATA_T_MONEY && (exp1->Types.Money.WholePart != exp2->Types.Money.WholePart || exp1->Types.Money.FractionPart != exp2->Types.Money.FractionPart))
+	if (!(exp1->Flags & EXPR_F_NULL) && exp1->DataType == DATA_T_MONEY && exp1->Types.Money.Value != exp2->Types.Money.Value)
```
sheesania, line 866: expCompareExpressionValues should also be tested for money type

centrallix/netdrivers/net_http_rest.c
sheesania: No, RestWriteAttrValue needs to tested for MoneyType json

centrallix/osdrivers/objdrv_dbl.c
sheesania: There should be tests for this, but they can probably be based off the objdrv_fp tests

centrallix/utility/json_util.c
sheesania: Currently still needs tests

centrallix/objectsystem/obj_datatypes.c
sheesania: @gbeeley As far as I can tell, `byteswap.h`/`bswap_64()` are part of the GNU C extensions. Is this safe to use without checking for its availability, or should I do some checking and conditional compilation like I did in #17?

gbeeley: Currently Centrallix uses the GNU C toolchain throughout.  We haven't done any work on building using LLVM or whatnot, and indeed ./configure checks for __GNUC__.  So, yes, it is acceptable to use GNU C extensions without preprocessor directives.

centrallix/netdrivers/net_http_rest.c
sheesania: 9eea437 is the closest we can get for now

centrallix/objectsystem/obj_datatypes.c
```diff
@@ -2474,3 +2473,28 @@ objDateAdd(pDateTime dt, int diff_sec, int diff_min, int diff_hr, int diff_day,
...
+    /**
+     * In the old representation, whole = floor($amount) and fraction = $amount - floor($amount). This means that
+     * fraction is always positive and always counts towards the total $amount in a positive direction.
+     * So if there is a negative with a non-zero fraction part...
+     **/
+    if (money.Value < 0 && money.Value%10000 != 0)
+    {
+        /** Add negative 1 then truncate fraction. (essentially floor() )**/
+        *wholePart = (money.Value - 10000)/10000ll;
+        /** Since fraction always counts away from zero, it is inverted**/
```
gbeeley, line 2492: Minor issue: this comment should probably be changed to "Since fraction always counts in the positive direction".

gbeeley: Non-blocking.

gbeeley (commented):
I've finished looking over the files here, next step is to actually work with it in testing.

sheesania (approved):

centrallix/expression/exp_evaluate.c
```diff
@@ -843,13 +773,7 @@ expEvalPlus(pExpression tree, pParamObjects objlist)
...
 	        objDataToMoney(i1->DataType, dptr, &m);
-		tree->Types.Money.WholePart = i0->Types.Money.WholePart + m.WholePart;
-		tree->Types.Money.FractionPart = i0->Types.Money.FractionPart + m.FractionPart;
-		if (tree->Types.Money.FractionPart >= 10000)
-		    {
-		    tree->Types.Money.FractionPart -= 10000;
-		    tree->Types.Money.WholePart++;
-		    }
+	        tree->Types.Money.Value = i0->Types.Money.Value + m.Value;
 		break;
```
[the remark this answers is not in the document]
sheesania: Okay, I opened #46 to keep track of that

gbeeley (approved):

PR #17, opened 2021-01-29
Centrallix testdriver only uses ncurses if it's available

### LINKED ISSUE #16: Centrallix testdriver should only try to use ncurses if it is available
See https://github.com/LightSys/centrallix/pull/14#discussion_r565674175

sheesania:
Depends on #14 
Fixes #16


gbeeley (approved):
Looks good - thanks!

Issue #25, opened 2021-03-05
Autoscaling/responsive widgets using CSS calc()


sheesania:
Currently, widgets are absolutely positioned and sized, which means they do not adapt to the size of the browser window or re-render when the window is resized. Implementing a truly responsive solution would be a lot of work given how much of the frontend currently leans on absolute positioning and sizing. So as a temporary stopgap, try using CSS calc() and percentages to at least resize widgets when the window is resized.

The calc function should look something like this:

`calc(initial width + (100% - container's initial width) * flex ratio)`

- "initial width" = width from first render (in px?)
- 100% = current width of window (as it's being resized)
- "container's initial width" = width of window from first render
- So "100% - containers initial width" = total new space available
- "flex ratio" = current widget's flexibility / total section flex (= % flex that is "mine")
- So `(100% - container's initial width) * flex ratio` is the amount of new space available that is "mine"

As far as I can tell, all positioning logic currently happens on the server side. `apos.c` uses the widget tree (`wgtr.c`) to calculate the x, y, width, height, flex width, and flex height of each "node" (widget?) (or maybe it creates "nodes"?), then builds and renders a grid that widgets place themselves in. The widgets decide how much of their (allotted?) width/height to use.

Supposedly every section and widget has "flex" area, and `apos.c` knows where inter-widget space needs to be static, vs what is flexible (possibly using section divisions?).

Lightning11wins:
> As far as I can tell, all positioning logic currently happens on the server side.

After working on this issue for months, I should note that, while this is mostly right, it's not 100% correct. Many widgets do client-side positioning logic, such as autolayout boxes (for automatic reflow logic), tab widgets (for tabs with automatic widths), pretty much all positioning for tables and their child DOM nodes is done client-side, and several others.

PR #26, opened 2021-03-05
Autoscale widgets with CSS calc() - primary work?

sheesania:
Part of #25 

As far as I can tell, this is where most of the work on the autoscaling project ended up?


centrallix/wgtr/wgtr.c
```diff
@@ -1515,7 +1518,8 @@ wgtrSetProperty(pWgtrNode widget, char* name, int datatype, pObjData val)
 pWgtrNode 
 wgtrNewNode(	char* name, char* type, pObjSession s,
 		int rx, int ry, int rwidth, int rheight,
-		int flx, int fly, int flwidth, int flheight)
+		int flx, int fly, int flwidth, int flheight,
```
Lightning11wins, line 1521: So we modify the function signature for wgtrNewNode() here, but this branch does not appear to modify any of the locations where the function is called. How do you intend for these new values to be generated?

PR #27, opened 2021-03-05
Autoscale widgets with CSS calc() - extra work?

sheesania:
Part of #25

This work is related to the autoscaling project, but I'm not sure precisely how.


Lightning11wins:
This work seems to be incomplete. Additionally, there appear to be several compiler errors that would need to be addressed. However, the concept of a "MainContainer" might be useful, but I can't find much explanation of what a "MainContainer" is, or why it should get special treatment in the layout. I'm likely just misunderstanding what the team was going for here, though.

Issue #29, opened 2021-03-05
Complete Centrallix SQL tests


sheesania:
CXSQL tests that still need to be completed:

- tests for remaining SQL functions and operators (summer 2020 was focused on string functions)
- tests for remaining SQL clauses -- currently, we have coverage of collections, joins, and the "on duplicate" clause for inserts. we'd need:
   - INSERT ... SELECT statements (more complete, there is some coverage with the on duplicate tests)
   - DELETE statements
   - UPDATE statements (including updating through a join)
   - SELECT with GROUP BY (including two-level/nested aggregates)
   - SELECT with ORDER BY
   - SELECT without a FROM (i.e. selecting constants)
   - query scope collections
   - declared objects, including setting and using declared object properties, on both application scope and query scope
   - subqueries

PR #30, opened 2021-03-05
Work on Centrallix SQL testsuite

sheesania:
Part of #29


gbeeley (approved):
These tests could use some additional review at some point, but things look fine to merge for now.

PR #39, opened 2021-04-14
qprintf %LL specifier

sheesania:


centrallix-lib/src/qprintf.c
```diff
@@ -58,47 +58,48 @@
 #define QPF_SPEC_T_DBL		(4)
 #define QPF_SPEC_T_NSTR		(5)
 #define QPF_SPEC_T_CHR		(6)
-#define QPF_SPEC_T_ENDSRC	(6)
+#define QPF_SPEC_T_ENDSRC	(7)
+#define QPF_SPEC_T_LL       (7)
```
gbeeley, line 62: Very minor: for clarity, the ENDSRC and LL defines should be swapped.

centrallix-lib/src/qprintf.c
```diff
@@ -1035,6 +1038,12 @@ qpfPrintf_va_internal(pQPSession s, char** str, size_t* size, qpf_grow_fn_t grow
 			    strval = tmpbuf;
 			    break;
 
+            case QPF_SPEC_T_LL:
+                llval = va_arg(ap, long long);
+                cplen = snprintf(tmpbuf, sizeof(tmpbuf), "%lld", llval);
```
gbeeley, line 1043: I can't think of any situation where a ≤ 0 return value would actually happen for %lld, but it'd be good form to check the return value from snprintf(), to ensure it is positive.

centrallix-lib/tests/test_qprintf_66.c
```diff
@@ -0,0 +1,51 @@
...
+    unsigned char buf[44];
+
+    *tname = "qprintf-66 %LL insertion in middle without overflow";
+    iter = 200000;
+    for(i=0;i<iter;i++)
+    {
+        buf[43] = '\n';
+        buf[42] = '\0';
+        buf[41] = 0xff;
+        buf[40] = '\0';
```
gbeeley, line 23: Recommend the trailing redzone be immediately adjacent to the area of buf[] that is written to.  I also think the first byte of the redzone should be nonzero, to detect a nul (end of qprintf'ed string) being written there.

gbeeley (changes requested):
Thanks!

centrallix-lib/src/qprintf.c
gbeeley: n/m it looks like the code already checks down below the switch.

gbeeley (approved):
Looks good - thanks!

PR #40, opened 2021-05-26
gitignore .vscode folder

sheesania:
Context for whether to include the .vscode folder in source control or not: https://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control I currently don't have any run/debug configurations, shared tasks, etc, so there is no need to share the .vscode folder.


gbeeley (approved):
Seems reasonable.  Thanks!

PR #44, opened 2021-07-12
Allow Centrallix C tests to be skipped over

sheesania:
For instance, say hypothetically that you're writing a test that uses the Sybase objectsystem driver, and should be skipped over if Sybase isn't enabled. The test could look something like this:

```
#include "config.h"
long long test(char** name)
{
    *name = "sybase driver test";

    #ifdef USE_SYBASE
    assert(sybaseThing());
    return 0;
    #else
    return 1;
    #endif
}
```

When USE_SYBASE is defined, the test driver will report that "sybase driver test" passes (if it actually passed, of course). When USE_SYBASE is not defined, the test driver will report that "sybase driver test" was skipped.

### DIFF
--- a/centrallix/tests/README
+++ b/centrallix/tests/README
@@ -82,8 +82,10 @@ NN is the subtest number, starting with 00 or 01, and {category} is a category
 name for the group of tests.
 
 The file should implement a function `long long test(**char name)` that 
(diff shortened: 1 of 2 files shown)


gbeeley (approved):
Great idea.

PR #49, opened 2022-05-09
Move or delete old/less relevant docs

sheesania:
This is obviously based on my own judgment of what would be more helpful to have out of the centrallix-sysdoc root. Feel free to let me know if you feel differently :)


gbeeley (changes requested):
Overall looks good, the ADAK PDF file however is current and should stay in the sysdoc folder.

sheesania:
@gbeeley My main concern with the ADAK PDF is that it's not directly relevant to the current implementation of Centrallix and there's an identical copy in the DistKeySim repository. Maybe a better idea would be to put in a "proposals" folder with the other sysdocs on to-be-implemented features?

gbeeley:
Yes, a "Proposed Features" folder would be a good idea for that document.  The security model documentation could also be moved there.

PR #50, opened 2022-05-09
Convert centrallix-sysdoc to using Markdown

sheesania:


sheesania:
@gbeeley Here are two first passes at converting the centrallix-sysdoc docs to Markdown. I wanted to give you a chance to see how they look before I blow through the rest of the docs.

I'm pretty much just:
- trying to convert the organization of the doc to use Markdown semantic elements like headers
- removing manual line breaks
- adding auto-generated tables of contents
- aiming for the finished doc to look good in plain text or as displayed by GitHub

gbeeley (changes requested):
Overall this looks good; it would be good to retain outline numbering however where it is present in the original document (1 2 3 ... and/or A B C ... and so forth).

centrallix-sysdoc/OSDriver_Overview.md
```diff
@@ -0,0 +1,296 @@
...
+                chain of open objects needed to get 
+                at the object the user wants.
+    3.	     O			Discovers that rootnode points to a local filesystem tree.
+    4.	     O -> F		OSML tries to get local filesystem driver to open the given path, /DirectoryOne/FileTwo.csv/rows/0
+    5.	     O <- F		Filesystem driver reports that it can only open /DirectoryOne/FileTwo.csv, and returns a handle for it.
+    6.	     O			Adds an object to the chain of open objects for the open file, and figures out that the content of the FileTwo.csv is CSV data.
+    7.	     O -> D		OSML tries to get the datafile driver to open up FileTwo.csv/rows/0, the remainder of the path.
+    8.	          D -> O	datafile driver uses the OSML API to access the open object created in step 5 above (the file).
+    9.	               O -> F	OSML requests content for the file from the local filesystem driver.
+    1.                 O <- F	Local filesystem driver returns the needed content.
```
gbeeley, line 180: Little oddity here -- the numbering has 1. instead of 10. and the O <- F isn't indented right.

centrallix-sysdoc/README
```diff
@@ -1,174 +1,104 @@
+# Intro
```
gbeeley, line 1: It looks like this needs to be renamed to README.md

gbeeley (changes requested):
Great work, we're very close here, I'm impressed re. how nice the tables look.  Just a couple of minor things to fix and this will be ready to merge.  I think there may be a reference to "JSON-REST.txt" in the kardia docs that we'll need to fix separately to refer to JSON-REST.md.

PR #54, opened 2022-05-19
Document research on ideas for making contributing to Centrallix easier

sheesania:
Depends on #51


gbeeley (approved):
This is all really good information.  I concur regarding the approach to testing.  Better unit testing would be ideal but would have to have been planned into the build and architecture much earlier on than where we are now.  If we ever port to another language like Rust or whatnot, that might be a good time to revisit this question.

PR #55, opened 2022-05-24
Duplicate checking

captain-nemo-10994:
Completed cosine compare function with letter pairs and added test suite.

### DIFF
--- a/centrallix/expression/exp_functions.c
+++ b/centrallix/expression/exp_functions.c
@@ -3937,35 +3937,7 @@ int exp_fn_lev_compare(pExpression tree, pParamObjects objlist, pExpression i0, 
 }
 
 // This is the size of the vector table. It is also used in calculating the table indices.
-const int VECTOR_TABLE_SIZE = 251;
-
-
-/*
- * exp_fn_i_trim_term
- * This method removes the puncuation and whitespace from a given string
- *
- * Paramters:
- * 	src : unprocessed string
- * 	dest : target string
- *
- * Returns: 
- * 	0 
- */
-int exp_fn_i_trim_term(char *src, char* dest)
-    {	
-    int i;
-    int j = 0;
-    for (i = 0; i < strlen(src)+1; i++)
-	{
-	if(!ispunct(src[i]) && !isspace(src[i]))
-	    {
-	    dest[j] = tolower(src[i]);
-	    j++;
-	    }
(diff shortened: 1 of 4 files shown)


centrallix/expression/exp_functions.c
```diff
@@ -3935,146 +3936,241 @@ int exp_fn_lev_compare(pExpression tree, pParamObjects objlist, pExpression i0,
...
+		temp += 75;
+		} 
+					
+	    int index = hash_char_pair(96, temp);
+	    table[index] += ((unsigned char)temp + 96) % 13;
+	    }
+	    // This case is for characters in the middle of the string. It adds no null character
+	    else if (i < strlen(term))
+		{
+		double temp1 = (int)term[i-1];
```
gbeeley, line 4031: Assuming you do eliminate the string copy (see other comment), you'll need a way of referencing the previously processed character.  I'd suggest just a temp variable storing what the previous char was, or your code for NUL if there was no previous char.

gbeeley (changes requested):
Thanks for your work!  A few tweaks and I think we're ready.

centrallix/expression/exp_functions.c
```diff
@@ -3935,146 +3936,230 @@ int exp_fn_lev_compare(pExpression tree, pParamObjects objlist, pExpression i0,
...
+	if (temp2 >= 48 && temp2 <= 57)
+	    {
+	    temp2 += 75;
+	    }
+
+	// Hash the character pair into an index
+	int index = exp_fn_i_hash_char_pair(temp1, temp2);
+
+	// Increment Frequency Table value by number from 0 to 13
+	table[index] += ((unsigned char)temp1 + (unsigned char)temp2) % 13;
```
gbeeley, line 4032: table[] holds type unsigned char, which depending on the string could wrap around with sufficient frequency additions.  Formerly the table[] values were doubles.  I'm wondering if the table data type should be a **short** (16 bit) anyhow.

Also, it would make sense to ensure that the table increment is nonzero.  So maybe n%13 + 1 instead of just n%13?

gbeeley (changes requested):
Two more items here - first is the frequency vector data type, which could wrap around, and second is the empty readme file in the tests/ directory.

gbeeley (approved):
Good job, guys!

PR #56, opened 2022-06-10
Duplicate checking

captain-nemo-10994:
Linked Argon2 library, added argon2id method in exp_functions.c, added test files for argon2id method


centrallix/expression/exp_functions.c
```diff
@@ -4159,6 +4161,88 @@ int exp_fn_cos_compare(pExpression tree, pParamObjects objlist, pExpression i0,
...
+ *	 0 if successful
+ *	 -1 if error
+ */
+int exp_fn_argon2id(pExpression tree, pParamObjects objlist, pExpression password, pExpression salt)
+{   
+    // Set the parameters for the Argon function that are not passed in
+    int HASHLEN = 32;
+    unsigned int T_COST = 2;
+    unsigned int M_COST = (1 << 16);
+    unsigned int PARALLELISM = 1;
```
gbeeley, line 4185: Is there some documentation on how these three values were chosen and what their typical impact is on the hash runtime and memory use?

(future direction: allow the 3rd and 4th params, optional, to specify time/mem costs; btw beyond the 3rd param you have to access it via tree->Children.)

gbeeley (changes requested):
Thanks!  Just need some info on how the constants were chosen.

centrallix/expression/exp_functions.c
captain-nemo-10994: To record what we talked about in person:
Those values were simply the defaults in the C example implementation. The actual values we use should be tuned to our particular system. 

I've added support for optional parameters with default values.

centrallix/expression/exp_functions.c
```diff
@@ -4159,6 +4161,92 @@ int exp_fn_cos_compare(pExpression tree, pParamObjects objlist, pExpression i0,
...
+    // Check for and set the optional parameters
+    // T_COST controls the time cost
+    // M_COST controls the memory cost
+    // PARALLELISM controls how many 'lanes' will be used
+    // HASHLEN is the size of the finished hash
+    unsigned int T_COST = (tree->Children.nItems >= 3)?((intptr_t)tree->Children.Items[2]):2;      // the default value should be changed according to system needs
+    unsigned int M_COST = (tree->Children.nItems >= 4)?((intptr_t)tree->Children.Items[3]):(1<<16);      // the default value should be changed according to system needs
+    unsigned int PARALLELISM = (tree->Children.nItems >= 5)?((intptr_t)tree->Children.Items[4]):1; // the default value should be changed according to system needs
+    unsigned int HASHLEN = (tree->Children.nItems >= 6)?((intptr_t)tree->Children.Items[5]):32;    // the default value should be changed according to system needs
+
```
gbeeley, line 4190: To prevent a denial of service condition, we should make sure there are bounds on acceptable values for these four parameters.  Lower bounds are whatever makes sense for valid values for argon2id.  Upper bounds:  maybe 1 second for time, 32MB for RAM, 8 threads for parallelism, and 256 bytes for hashlen?  We can always fix those later if those limits aren't appropriate.

For a later project, it might be good to only allow a user to have one argon() running at a time system-wide, to prevent a user from bogging down / DoSing the server with argon() calls since it is intentionally an expensive function.  But that would require adding a mutex / semaphore to a session somewhere, and I'd have to track down where'd be the best place to put that.

centrallix/tests/test_argon2id_00.to
```diff
@@ -0,0 +1,12 @@
...
+query select 'case   1' = argon2id("password", "thisisasalt")
+query select 'case   2' = argon2id("password", "differentsalt")
+query select 'case   3' = argon2id("awayhaulaway", "wellhaulawayjoe")
+query select 'case   4' = argon2id("aaaaaaaaaaaaaaaa", "aaaaaaaaaaaaaaaa")
+query select 'case   5' = argon2id("P455w0rdW1th-D1ff3r3n5t!Ch@r@ct3r5??", "sodiumchloride")
+query select 'case   6' = argon2id("111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111", "111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111")
+query select 'case   7' = argon2id(":^:#@)$!#@(*&!#)$", "!@)*#&!@(*#&)!(*@&#)(!*@&#)(*")
+query select 'case   8' = argon2id("", "thissalthasanemptypassword")
+query select 'case   9' = argon2id("        ", "spacethefinalfrontier")
+query select 'case  10' = argon2id("		", "putitonmytab")
```
gbeeley, line 12: Let's add some tests for the new complexity parameters, and let's add some tests for corner cases like null parameters (null/nonexistent complexity params should force a default value, but null data or salt should return null).

Also if we could include a couple of the standard tests from https://github.com/P-H-C/phc-winner-argon2/blob/master/src/test.c (see the last group of tests for argon2id) that would be great.  Note that those test calls generate multiple outputs, only one of which is the standard hexstring output.

gbeeley (changes requested):
Thanks!  A couple more changes to make, including tests for the new params.

centrallix/expression/exp_functions.c
captain-nemo-10994: Sounds good. That's easy to add.

centrallix/tests/test_argon2id_00.to
captain-nemo-10994: Ah, okay. I had simply been using `return -1` and putting up an error message when null password or null salt occurs. To clarify: you're saying you'd like me to store a null value in tree->String when that happens?

centrallix/expression/exp_functions.c
gbeeley: I should add -- the above code doesn't look correct either.  The integer would be ((pExpression)tree->Children.Items[3])->Integer, but you'd also need to verify the datatype is DATA_T_INTEGER and the flags don't contain EXPR_F_NULL.

centrallix/tests/test_argon2id_00.to
gbeeley: If the data or salt is not supplied at all, that's an error condition.  If either are null, go ahead and return null, as that is typical for SQL functions.  To give a null result, just set tree->DataType to DATA_T_STRING and tree->Flags to include EXPR_F_NULL, and then return 0.

captain-nemo-10994: Got it. Thanks!

centrallix/expression/exp_functions.c
captain-nemo-10994: Okay. I'll make those changes as well.

centrallix/expression/exp_functions.c
```diff
@@ -4159,6 +4161,91 @@ int exp_fn_cos_compare(pExpression tree, pParamObjects objlist, pExpression i0,
...
+    //check password and salt
+    if (!password || !salt)
+	{
+	mssError(1, "EXP", "Invalid Parameters: Method requires two arguments");
+	return -1;
+	}
+    else if ((password->Flags | salt->Flags) & EXPR_F_NULL)
+	{
+	tree->Flags |= EXPR_F_NULL;
```
gbeeley, line 4190: Also set data type here before returning (DATA_T_STRING).

centrallix/expression/exp_functions.c
```diff
@@ -4159,6 +4161,91 @@ int exp_fn_cos_compare(pExpression tree, pParamObjects objlist, pExpression i0,
...
+    unsigned int M_COST = ((tree->Children.nItems >= 4) &&  ((tree->Children.Items[3]) != NULL) && ((pExpression)tree->Children.Items[3])->DataType == DATA_T_INTEGER && ((pExpression)tree->Children.Items[3])->Integer < INT_MAX && ((pExpression)tree->Children.Items[3])->Integer >= 64) ? ((pExpression)tree->Children.Items[3])->Integer:(1<<16);
+    // PARALLELISM determines the number of threads or 'lanes' used by the algorithm
+    unsigned int PARALLELISM = ((tree->Children.nItems) >= 5 && ((tree->Children.Items[4]) != NULL) && ((pExpression)tree->Children.Items[4])->DataType == DATA_T_INTEGER &&((pExpression)tree->Children.Items[4])->Integer <= 8 && ((pExpression)tree->Children.Items[4])->Integer >=1) ? ((pExpression)tree->Children.Items[4])->Integer:1;
+    // HASHLEN is the size of the finished hash
+    unsigned int HASHLEN = ((tree->Children.nItems >= 6) && ((tree->Children.Items[5]) != NULL) && ((pExpression)tree->Children.Items[5])->DataType == DATA_T_INTEGER && ((pExpression)tree->Children.Items[5])->Integer < 256 && ((pExpression)tree->Children.Items[5])->Integer >= 4) ? ((pExpression)tree->Children.Items[5])->Integer:32;
+
```
gbeeley, line 4203: These four items should be rewritten as more traditional if-blocks for readability, or else write a little default-setting function that you can call which simplifies how this code reads.

centrallix/expression/exp_functions.c
```diff
@@ -4159,6 +4161,91 @@ int exp_fn_cos_compare(pExpression tree, pParamObjects objlist, pExpression i0,
...
+	mssError(1, "EXP", "Salt is too short: Salt must be at least 8 bytes (16 recommended)");
+	return -1;
+	}
+    // this call to the argon2id_hash_raw method is where the magic happens
+    // after this call, the hashed password is written in hashvalue   
+    argon2id_hash_raw(T_COST, M_COST, PARALLELISM, pwd, pwdlen, slt, sltlen, hashvalue, HASHLEN);
+    
+    // this is where we write the contents of hashvalue to tree using qpfPrintf
+    if (HASHLEN*2+1 > 63)
+	{
```
gbeeley, line 4232: Instead of using the constant 63 here, use sizeof(tree->Types.StringBuf) - 1.  For example:
`if (HASHLEN*2+1 > sizeof(tree->Types.StringBuf))`
Since the nul terminator is already accounted for in the +1 on the left hand side, we don't have to subtract one from the length of the buffer.

centrallix/expression/exp_functions.c
```diff
@@ -4159,6 +4161,91 @@ int exp_fn_cos_compare(pExpression tree, pParamObjects objlist, pExpression i0,
...
+	tree->Flags |= EXPR_F_NULL;
+	return 0;
+	}
+
+    // The default values of the following four variables should be tuned for each specific system's needs
+    // T_COST determines the number of passes the algorithm makes
+    unsigned int T_COST = ((tree->Children.nItems >= 3) && ((tree->Children.Items[2]) != NULL) && ((pExpression)tree->Children.Items[2])->DataType == DATA_T_INTEGER && ((pExpression)tree->Children.Items[2])->Integer < 8 && ((pExpression)tree->Children.Items[2])->Integer > 0) ? ((pExpression)tree->Children.Items[2])->Integer:2;
+    // M_COST determines the amount of memory cost in kilobytes
+    unsigned int M_COST = ((tree->Children.nItems >= 4) &&  ((tree->Children.Items[3]) != NULL) && ((pExpression)tree->Children.Items[3])->DataType == DATA_T_INTEGER && ((pExpression)tree->Children.Items[3])->Integer < INT_MAX && ((pExpression)tree->Children.Items[3])->Integer >= 64) ? ((pExpression)tree->Children.Items[3])->Integer:(1<<16);
+    // PARALLELISM determines the number of threads or 'lanes' used by the algorithm
```
gbeeley, line 4199: We'll need to limit the memory cost much lower than INT_MAX, which would be 2TB.

gbeeley (changes requested):
Getting close!  Thanks for your work on this.

centrallix/expression/exp_functions.c
captain-nemo-10994: Got it!

centrallix/expression/exp_functions.c
captain-nemo-10994: I went with the if-blocks.

centrallix/expression/exp_functions.c
captain-nemo-10994: Ah, I messed up some conversions. Whoops. That's fixed now.

centrallix/expression/exp_functions.c
```diff
@@ -4159,6 +4161,127 @@ int exp_fn_cos_compare(pExpression tree, pParamObjects objlist, pExpression i0,
...
+	return 0;
+	}
+
+    // The default values of the following four variables should be tuned for each specific system's needs
+    // T_COST determines the number of passes the algorithm makes
+    unsigned int T_COST = 2; 
+    if ((tree->Children.nItems >= 3) && 
+	((tree->Children.Items[2]) != NULL) && 
+	((pExpression)tree->Children.Items[2])->DataType == DATA_T_INTEGER && 
+	((pExpression)tree->Children.Items[2])->Integer < 8 &&
```
gbeeley, line 4201: Is 8 the number of iterations?  If so I'd guess this would need to be much higher, so we can cause the algorithm to take up to about a second to run.

gbeeley (changes requested):
One question about T_COST (iterations).  Thanks!

centrallix/expression/exp_functions.c
captain-nemo-10994: The upper bound for T_COST is now set to 24. I consistently got approximately a 1 second execution time with 12 iterations.

gbeeley (approved):

PR #58, opened 2022-06-16
Css efficiency

captain-nemo-10994:
I've edited a handful of the files based on the instructions in your email. I feel like I'm missing something. I don't think I quite understand this paragraph from the email:

> "You'll also have to add classes to DIVs and other elements that you are
> writing the tag.class rules for.  The htrAddBodyItemLayer_va() call in
> htdrv_image.c gives an example. In that case, there is a DIV with an IMG
> inside of it.  The class on the DIV is provided by the 5th param to
> htrAddBodyItemLayer_va().  The class for the IMG is provided in the text
> containing the tag."

Could you clarify this any further?

### DIFF
--- a/centrallix/htmlgen/htdrv_chart.c
+++ b/centrallix/htmlgen/htdrv_chart.c
@@ -51,6 +51,13 @@ static struct
     int idcnt;
 } HTCHT;
 
+
+int htchtSetup(pHtSession s)
+	{
+	htrAddStylesheetItem(s, "	div.wcht { POSITION:absolute; VISIBILITY:inherit; }\n");
+	}
+
+
 int
 htchtCheckBrowserSupport(pHtSession session)
     {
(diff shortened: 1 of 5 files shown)


gbeeley (changes requested):
Hi captain-nemo-10994, github won't let me comment on the specific lines of code you didn't (yet) change, so I'll put the comments here.  "Classes" can be applied to HTML elements, see https://www.w3schools.com/html/html_classes.asp for an explanation.  In order to use classes in CSS rules (such as tagname.classname), you need to apply the classname in question to the tags you want the rule to apply to.  So, for example  or <div class="overlaydiv">content goes here</div>.  See the example code in htdrv_image.c lines 164 and 165.  The htrAddBodyItemLayer_va() function there is a convenience function that creates a <div> or <iframe>.  It's defined in ht_render.c.

PR #60, opened 2022-07-08
Cred mgr test

nboard:
This code should achieve the following: 
- Test and confirm the following Crypto functionality:
1. Init/cleanup
2. gen salt, iv, and random keys
3. RSA keypair generation
4. AES and RSA encryption/decryption 
5. Key destruction 

- Test and confirm the following DB functionality:
1. init/clode
2. insert, retrieve, update, and delete for all three tables (auth, resc, and user data)
3. delete all for resc and auth
4. retrieve linked list for auth

- Provide the following DB changes (along with tests):
1. Remove salt from the Auth table
2. Move teh AuthClass field from the resc table to the auth table
3. Change the group retrieve and delete for auth to use the AuthClass field 
4. Fix return codes for updates and deletes that affect 0 items
5. Individual updates and deletes for auth table, as well as retrieval by PK

Please note that the credentials_mgr.h and credentials_mgr.c documents were changed to accommodate changes in the way functions are called. The changed regions did not appear to be used anywhere, so the changes should not break anything.


centrallix/tests/test_cxss-credDB_00.c
```diff
@@ -0,0 +1,20 @@
...
+#include "cxss/credentials_db.h"
+
+long long
+test(char** name)
+    {
+     *name = "CXSS Cred DB 00: Basic Init";
+
+    
+    /** Basic test of init and release */
+    char * PATH = "/home/devel/test.db";
```
gbeeley, line 12: Be sure to not have an install-dependent path in the tests.  There may not always be a user "devel".  You may want to use a relative path in the current folder instead of an absolute path.

gbeeley (changes requested):
Looking good!  A few changes needed here.

centrallix/tests/test_cxss-crypto_00.c
```diff
@@ -0,0 +1,16 @@
...
+
+long long
+test(char** name)
+    {
+     *name = "CXSS Crypto 00: Basic Init";
+
+    /** Basic test of init and release */
+    cxss_internal_InitEntropy(1280); /* must be into first */
+    cxssCryptoInit();
+    cxssCryptoCleanup();
```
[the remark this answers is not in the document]
nboard: This was done just to ensure it does not segfault. There does not appear to be anything done by these functions that can be easily confirmed externally.

centrallix/tests/test_cxss-crypto_05.c
```diff
@@ -0,0 +1,117 @@
...
+    assert(strcmp(plaintext, plaintext2) == 0);
+    assert(strcmp(plaintext, short1) == 0);
+
+
+    cxssCryptoCleanup();
+
+    return 0;
+    }
+
+void clearBuff(char* array, int length)
```
gbeeley, line 111: You can use memset() or bzero() for this purpose.

gbeeley (changes requested):
One minor change.  Thanks!

centrallix/tests/test_cxss-crypto_05.c
nboard: Got it! thanks for the tip

gbeeley (approved):
Looks good.  Thanks Noah!

nboard:
I think I did not realize the approval meant I should have merged it. Is it fine to merge now, or does that need reevaluated now that is has been so long?

PR #61, opened 2022-07-28
Writeup of pros, cons, and options for using React in Centrallix

sheesania:


gbeeley (approved):
Looks good!  Thank you.

PR #62, opened 2023-09-20
Char encode attr

nboard:
This should be ready to be merged in. It ended up fitting in relatively smoothly with the rest of the code.

### DIFF
--- a/centrallix/osdrivers/objdrv_http.c
+++ b/centrallix/osdrivers/objdrv_http.c
@@ -1729,13 +1729,11 @@ http_internal_GetPageStream(pHttpData inf)
 		    }
 		else if(inf->ExpectedContentCharset)
 		    inf->ContentCharset = nmSysStrdup(inf->ExpectedContentCharset);
-printf("******** HTTP: set content based on header: %s\n\n", inf->ContentCharset);
 		}
 	    else
 		{
 		inf->ContentType = nmSysStrdup(inf->RestrictContentType);
 		inf->ContentCharset = nmSysStrdup(inf->ExpectedContentCharset);
-printf("******** HTTP: no content type in header; set based on expected: %s\n\n", inf->ContentCharset);
 		}
 
 	    /** Make sure the content type is allowed **/
@@ -2163,7 +2161,6 @@ httpOpen(pObject obj, int mask, pContentType systype, char* usrtype, pObjTrxTree
 	inf->OverrideContentCharset = strcmp(http_internal_GetConfigString(inf, "override_content_charset", "false"), "true") == 0;
 	inf->Protocol = http_internal_GetConfigString(inf, "protocol", "http");
 	inf->Cipherlist = http_internal_GetConfigString(inf, "ssl_cipherlist", "");
-printf("\n******** HTTP: reading config - %s, %d\n\n", inf->ExpectedContentCharset, inf->OverrideContentCharset);
 
 	/** Valid method? **/
 	if (strcmp(inf->Method, "GET") && strcmp(inf->Method, "POST") && strcmp(inf->Method, "DELETE"))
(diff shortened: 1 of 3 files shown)


centrallix/osdrivers/objdrv_http.c
```diff
@@ -1712,10 +1719,21 @@ http_internal_GetPageStream(pHttpData inf)
...
+		    if (strchr(inf->ContentCharset, ';'))
+			*(strchr(inf->ContentCharset, ';')) = '\0';
+		    }
+		else if(inf->ExpectedContentCharset)
+		    inf->ContentCharset = nmSysStrdup(inf->ExpectedContentCharset);
 		}
 	    else
 		{
 		inf->ContentType = nmSysStrdup(inf->RestrictContentType);
+		inf->ContentCharset = nmSysStrdup(inf->ExpectedContentCharset);
```
gbeeley, line 1736: nmSysStrdup() will not accept a NULL input, which could happen if that property in the http object is unset.  Similar issue a few lines above.

gbeeley (changes requested):
A couple issues to fix.  Thanks!

centrallix/osdrivers/objdrv_http.c
nboard: I mentioned before I thought I had tested it, but in order to get here there had to be no content type field in the response header, which was not a case I tested after all. 

I am not sure what is wrong with the one a few lines above; the other two usages of nmSysStrdup I added already had null checks?

centrallix/osdrivers/objdrv_http.c
```diff
@@ -1712,10 +1719,22 @@ http_internal_GetPageStream(pHttpData inf)
...
+		/* add charset if present */
+		char * charsetStart = NULL;
+		if((charsetStart = strstr(hdr_val, "charset=")) != NULL && !inf->OverrideContentCharset)
+		    {
+		    inf->ContentCharset = nmSysStrdup(charsetStart+8);
+		    if (strchr(inf->ContentCharset, ';'))
+			*(strchr(inf->ContentCharset, ';')) = '\0';
+		    }
+		else if(inf->ExpectedContentCharset)
+		    inf->ContentCharset = nmSysStrdup(inf->ExpectedContentCharset);
```
nboard, line 1731: Is this the line you were referring to? The way the line wraps on my screen makes it harder to notice it is actually inside the NULL check on the line 1730

nboard:
Regarding the force push: I started to add one of the temporary fixes here, but I realized it both interfered with this pull request, and would also work better as the start of its own feature branch.

nboard:
Merged in master to make passing of to disciple makers for some testing simpler. The merge did not cause any conflicts, so the pull request should be in more or less the same state as before.

nboard:
The xml_cache_fix branch should be done as well now. It is based off of this branch, so we could merge the xml_cache_fix branch into this one and do one pull into master, or finish this pull request and then open a separate pull request for the xml_cache_fix  branch.

centrallix/osdrivers/objdrv_xml.c
```diff
@@ -490,6 +490,12 @@ xml_internal_ReadDoc(pObject obj)
 	    /** parse the document **/
 	    ptr=(char*)malloc(XML_BLOCK_SIZE);
 	    ctxt=xmlCreatePushParserCtxt(NULL,NULL,NULL,0,"unknown");
+	    /** change the encoding if needed*/
+	    ObjData contentCharset;
+	    objGetAttrValue(obj->Prev, "content_charset", DATA_T_STRING, &contentCharset);
+	    if(contentCharset.String != NULL)
```
gbeeley, line 496: We need to also check the return value of objGetAttrValue().  If it returns -1 then it failed to get the property, in which case contentCharset.String is unchanged and holds whatever uninitialized value was in it when it was declared.  A return value of -1 should be handled the same way a NULL value is handled in this case.

gbeeley (changes requested):
Just one thing to fix here, then let's merge this first and afterward do a PR for the xml caching branch.

centrallix/osdrivers/objdrv_xml.c
nboard: I committed the requested changes. Because the XML cache bug still exists in this branch testing was not as thorough as I would have liked, but this change should not have been able to change much anyway

gbeeley (approved):
Thank you nboard!

PR #63, opened 2023-11-17
Xml cache fix

nboard:
I merged in master and tested everything. Things seem to be working as expected (which also further confirms the char_encode_attr branch is working as expected).

### DIFF
--- a/centrallix/osdrivers/objdrv_xml.c
+++ b/centrallix/osdrivers/objdrv_xml.c
@@ -448,15 +448,20 @@ xml_internal_ReadDoc(pObject obj)
 	/** Determine path of just the XML file itself **/
 	path=obj_internal_PathPart(obj->Pathname,0,obj->SubPtr);
 
+/** TODO: use a full path including params for the cache check **/
+//char hash[4 + OBJSYS_MAX_PATH*2];
+//obj_internal_GetDCHash(obj->Pathname, obj->Mode, hash, (4 + OBJSYS_MAX_PATH*2), obj->SubPtr);
+
 	/** Check cache for an existing copy already **/
 	if((pCache=(pXmlCacheObj)xhLookup(&XML_INF.cache, path)))
 	    {
 	    if(XML_DEBUG) printf("found %s in cache\n", path);
 
 	    /** found match in cache -- check modification time **/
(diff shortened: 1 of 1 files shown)


centrallix/osdrivers/objdrv_xml.c
```diff
@@ -469,13 +471,13 @@ xml_internal_ReadDoc(pObject obj)
...
 	    {
-	    if(XML_DEBUG) printf("couldn't find %s in cache\n",path);
+	    if(XML_DEBUG) printf("couldn't find %s in cache\n",hash);
 	    pCache=(pXmlCacheObj)nmMalloc(sizeof(XmlCacheObj));
 	    if(!pCache) return NULL;
 	    memset(pCache,0,sizeof(XmlCacheObj));
 	    if (objGetAttrValue(obj->Prev, "last_modification", DATA_T_DATETIME, POD(&pDT)) == 0)
 		pCache->lastmod.Value = pDT->Value;
-	    strtcpy(pCache->Pathname, path, sizeof(pCache->Pathname));
+	    strtcpy(pCache->Pathname, hash, sizeof(pCache->Pathname));
```
gbeeley, line 480: Ensure that pXmlCacheObj -> Pathname is the same length as 'hash', otherwise you could get false cache misses because the cache's idea of the serialization is different from the lookup computed value.  Perhaps use a #define in order to ensure that those lengths match.

gbeeley (changes requested):
One change needed re. serialization ('hash') length in the cache vs in the lookup.  Thanks! :)

centrallix/osdrivers/objdrv_xml.c
nboard: I think this latest change should address the problem. Let me know what you think

gbeeley: A couple of minor notes (not blocking this PR) - in the GetDCHash call, use sizeof(hash) rather than XML_HASH_SIZE, and in the #define where you have a computed value, put parens around the expression so you don't get unexpected interactions such as 2*XML_HASH_SIZE (hypothetical) not doing what one might expect.

gbeeley (approved):
Looks good - thanks nboard!

PR #64, opened 2023-11-27
Mysql case compare

nboard:
I added the necessary collation changes for upper and lower, and the changes seem to be working as expected.

### DIFF
--- a/centrallix/osdrivers/objdrv_mysql.c
+++ b/centrallix/osdrivers/objdrv_mysql.c
@@ -68,7 +68,7 @@ typedef struct
     char        Password[64];
     char        DefaultPassword[64];
     char        Database[MYSD_NAME_LEN];
-    char        DatabaseCharset[64];
+    char        DatabaseCollation[64];
     char        AnnotTable[MYSD_NAME_LEN];
     char        Description[256];
     int                MaxConn;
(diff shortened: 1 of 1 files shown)


centrallix/osdrivers/objdrv_mysql.c
```diff
@@ -344,6 +345,22 @@ mysd_internal_GetConn(pMysdNode node)
...
+	if (!result || result == MYSD_RUNQUERY_ERROR){
+	    /** just set to all NULLs **/
+	    memset(node->DatabaseCharset, '\x00', sizeof(node->DatabaseCharset));
+	} else {
+	    /** collect the character name **/
+	    MYSQL_ROW row = mysql_fetch_row(result);
+	    if(row == 0){
+		memset(node->DatabaseCharset, '\x00', sizeof(node->DatabaseCharset));
+	    } else {
+		strncpy(node->DatabaseCharset, row[0], sizeof(node->DatabaseCharset));
```
gbeeley, line 359: Use strtcpy() here instead.  strncpy() does not null-terminate if the copy uses the entire destination buffer.

centrallix/osdrivers/objdrv_mysql.c
```diff
@@ -1832,6 +1849,49 @@ mysd_internal_TreeToClause(pExpression tree, pMysdTable *tdata, pXString where_c
...
 		    xsConcatenate(where_clause, ") ", 2);
 		    }
+		else if (!strcmp(tree->Name,"upper"))
+		    {
+		    /** check if database is using an expected character set **/
+		    char* collation = NULL; 
+		    pMysdTable tempTdata = *tdata;
+
+		    if(strstr(tempTdata->Node->DatabaseCharset, "latin1") != NULL) collation = "latin1_bin";
+		    else if (strstr(tempTdata->Node->DatabaseCharset, "utf8mb4") != NULL) collation = "utf8mb4_bin";
```
gbeeley, line 1859: Do we also need to handle the utf8mb3 case here?  Any other cases?

(same for lower() below)

The upper/lower collation setting could also be detected and set during charset detection in GetConn(), so we don't need to have the logic twice here in the function special cases.

gbeeley (changes requested):
A couple items to fix - thank you!

centrallix/osdrivers/objdrv_mysql.c
nboard: Switched them out. Thanks for the help!

centrallix/osdrivers/objdrv_mysql.c
nboard: I am unsure what collations would be useful/necessary, but most of the character sets listed as supported by MariaDB have a collation that is of the form <CHARSET NAME>_bin, by appending _bin to the end of the character set name we support almost everything. The only exceptions are the binary charset, which I added a check for, and the uca1400 collations, which do not appear to be supported by the version of MariaDB we use and have a character set of NULL; it is treated like an empty/failed request.

Good idea; I move the logic and store the collation rather than the character set now.

PR #65, opened 2024-02-12
Http xml memory fix

nboard:
I fixed the major memory leak found in the HTTP driver, and also a minor memory leak found in the expression logic. 
I also included changes the XML driver converting the cache  from a hash-map to a hash-queue. Although this had far less effect on the memory usage than the HTTP driver's memory leak, I did notice that the memory usage now does not grow noticeably, whereas with just the leak fix it would still climb slowly.

### DIFF
--- a/centrallix/osdrivers/objdrv_xml.c
+++ b/centrallix/osdrivers/objdrv_xml.c
@@ -214,7 +214,7 @@ typedef struct
 /*** GLOBALS ***/
 struct
     {
-    XHashTable	cache;
+    XHashQueue	cache;
     regex_t namematch;
     }
     XML_INF;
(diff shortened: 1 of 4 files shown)


nboard:
I caught a mistake in the XML driver changes, and fixed another minor leak in the expression functions.

nboard:
I noticed it was behind master. There were no conflicts, but I went ahead and merged master into it

centrallix/osdrivers/objdrv_xml.c
```diff
@@ -214,7 +214,7 @@ typedef struct
 /*** GLOBALS ***/
 struct
     {
-    XHashTable	cache;
+    XHashQueue	cache;
```
gbeeley, line 217: Should also #include "cxlib/xhashqueue.h" with the other includes at the beginning of the file, mainly for good hygiene.  The file is already included from obj.h, which is why things are working.

gbeeley (changes requested):
One minor fix needed - thanks!

centrallix/osdrivers/objdrv_xml.c
nboard: Noted

gbeeley (approved):
Looks good - thanks!

PR #68, opened 2024-11-27
Dynamic chart series

nboard:
Added functionality to the report driver to support charts with a dynamic number of series. Line graphs and bar graphs now support legends.


centrallix-doc/Report/chart.html
```diff
@@ -109,6 +109,16 @@ <h3 class="childproperties">Child Properties:</h3>
 <td class="description">The color to be used for the series.</td>
 </tr>
 <tr>
+<td class="name">legend_name</td>
```
gbeeley, line 112: These .html files are generated from the report.xml data file - you'll want to edit the xml file and then re-generate the various other files using the Makefile.  Note that using the Makefile will overwrite these two .html files.

centrallix-os/samples/chart_auto.rpt
```diff
@@ -0,0 +1,109 @@
...
+	    show_value=yes;
+	    show_legend=yes;
+
+	    chart_x "report/chart-axis" { axis=x; label="place"; }
+
+	    chart_y "report/chart-axis" { axis=y; label="value"; }
+
+	    x_value=runserver(:values_qy:x);
+	    y_value=runserver(:values_qy:y);
+	    series_value=runserver(:values_qy:key);
```
gbeeley, line 62: Can we combine report/auto-chart with report/chart, and have the functionality split if series_value is provided (not null)?  The user would need to either provide a series_value or provide one or more series objects.

centrallix-os/samples/chart_auto.rpt
```diff
@@ -0,0 +1,109 @@
...
+	    }
```
gbeeley, line 63: Can we also have a series_label property for display in the legend?  Sometimes the value will be a code or number but the label could then be human-readable.  Not sure what all MathGL allows.

centrallix/osdrivers/objdrv_report_v3.c
```diff
@@ -4143,6 +4240,509 @@ rpt_internal_DoChart(pRptData inf, pStructInf chart, pRptSession rs, int contain
...
+	    goto error;
+	    }
+
+	/** Set up rendering context **/
+	ctx = (pRptChartContext)nmMalloc(sizeof(RptChartContext));
+	if (!ctx)
+	    goto error;
+	memset(ctx, 0, sizeof(RptChartContext));
+	ctx->inf = inf;
+	rpt_internal_GetString(inf, chart, "show_legend", &ptr, "no", 0);
```
gbeeley, line 4323: This can use rpt_internal_GetBool().

centrallix-os/samples/chart_auto.rpt
nboard: I managed to combine the functions - there was more redundancy than I remembered, and most of the the conflicts could be resolved with simple if/else statements. I did split the actual reading of the values out into 2 new functions though - the auto-chart version was a couple hundred lines and it felt more readable that way.

centrallix-os/samples/chart_auto.rpt
nboard: I had already added the capacity for normal charts to have a custom legend name (since the name of the series object in .rpt files would seldom work well), so adding it for auto charts turned out to be a relatively simple addition.

centrallix/osdrivers/objdrv_report_v3.c
nboard: That makes the code simpler - thank you!

centrallix-doc/Report/chart.html
nboard: The documentation should now reflect the latest functionality and is entered into the report.xml file.

nboard:
I believe I have addressed all of the issues. Please review my fixes and let me know if anything else needs changed. Thanks!

centrallix-doc/Report/report.xml
```diff
@@ -343,14 +344,20 @@ myTable "report/table"
 		<properties>
 			<property name="box" type="yes/no">Whether to draw a box around the chart's data area (the box is drawn just around the graph or chart itself, not around the axis and tick labels and titles).  Default "no".</property>
 			<property name="chart_type" type="string">The type of chart:  bar, line, or pie.</property>
-			<property name="color" type="string">The default color to be used for the chart.</property>
+			<property name="color" type="string">The default color to be used for the chart. If nothing is specified elsewhere, all series will be made this color. If a palette is also specified, this color will replace any invalid colors.</property>
+			<property name="palette" type="string">A comma separated list of colors to be used by charts with dynamic series. This takes priority over the color property and the default palette. Once the palette runs out of to map to series, it loops through the colors again. Example: "green,,#014421", would have series colored a bright green, the default, and then a dark green.</property>
```
gbeeley, line 348: For the `palette` property, I'd recommend doing a structure-file-parsed comma separated list instead of a comma separated list inside a string.  That way, individual elements of the list are easier to specify with an expression or whatnot.  For example:

`palette = "green", "red", "blue";`

instead of

`palette = "green, red, blue";`

The updated list can be retrieved just using rpt_internal_GetString() calls with successively higher index (nval) values.  Just use a NULL default value and stop iterating when a NULL comes back.

gbeeley (changes requested):
One small change to palette settings.  Thanks!

centrallix-doc/Report/report.xml
gbeeley: Continuing to noodle on this -- looking ahead, if we want to have a system-wide chart palette preference, we'd need a way to SELECT a list of colors from a preference table.  I'd just need to make sure the expression language and report writer could handle a list created by a subquery expression.  On the other hand, in the short term, it might actually be useful for the report writer code to handle either of the above cases - parsing colors in a list from a single string vs. having them in separate strings.

nboard: Okay! The driver can now handle any mixture of lists of strings and strings containing lists.

gbeeley (approved):
Looks good - thanks nboard!

PR #69, opened 2025-02-01
Fixed bug where jumping to the end of a long table cleared it.

nboard:


gbeeley (approved):
Looks good - thanks nboard!

PR #70, opened 2025-03-06
Improve Non-Paginated HTML Report Formatting (Code-A-Thon 2025)

brennenputh:
This pull request improves upon the HTML report format already present in Centrallix by replacing table-based HTML with flexbox-based HTML.  The goal is to generate a HTML document similar in format and style to the PDF report using CSS elements of recent browser support.

Code-A-Thon Team: Brennen Puth, Levi Larsen, Kai Delsing, Josiah Hardaway


brennenputh:
All objects which were seen were revamped, particularly strings, tables, images, areas, and the root layer.
If this subsystem is to be used for HTML emails in the future, it would be necessary to move back towards table layouts.

RubberDuckCoding:
Creation of the function `prt_htmlfm_GetFont` should be reverted. To my knowledge it is not C-safe to return a string from a function (as you are returning a char * to the stack space that has just been released as the function returns).

brennenputh:
I had accidentally unchecked the "allow maintainers to edit" box it seems, and I just noticed, my bad.  It's allowed now, and ya'll should be able to push to my branch to resolve these things.

gbeeley:
@brennenputh - thanks!  I think we'll likely end up moving your work to a new branch in the LightSys copy before merging into master, though.  We're needing to separate reformatting and actual code into separate commits, which unfortunately would lose the current commit history with this PR, but in the long run it'll help us see things more clearly with git blame (e.g., having git blame ignore reformatting-only revs).  @RubberDuckCoding is a former Code-a-Thon-er who is working on finishing this project & getting it working well in Gmail/etc.

gbeeley:
Here's another reference that can be helpful on C code safety.  There's things we can all learn from this (even for me, who contributed to this effort some years ago):  https://resources.sei.cmu.edu/downloads/secure-coding/assets/sei-cert-c-coding-standard-2016-v01.pdf

PR #72, opened 2025-06-20
Minor Improvements

Lightning11wins:
While reading some docs, I made a few minor improvements. These only improve the docs by about 5-10%, so they might not even be worth the time to review. Still, I'll leave this here in case we decide to use it at some point.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.idea/*
 .depend
 Makefile
 config.log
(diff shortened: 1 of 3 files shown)


Lightning11wins:
I wanted to make sure that devs using IntelliJ IDEA didn't accidentally push their editor files. I doubt most of us are using that editor, but it was an easy change to make.

Lightning11wins:
This looks like fine, usable work. It's not at all a large PR, but it's still a small improvement on the pieces of documentation that it improves.

gbeeley (approved):
Looks good - thanks

PR #89, opened 2026-03-04
Add AI-generated HTML

Lightning11wins:
_Changes extracted from #85._
No logic affected. Minimal review needed (skipping Greptile).


Lightning11wins:
I'm marking this as trivial because we don't need to verify much of anything about it. This is just example HTML so that I have something to point the HTML widget at, so as long as it can render in a browser, that's basically all we need.

PR #92, opened 2026-03-11
Fix Potential Buffer Overflow

Lightning11wins:
Fix a potential buffer overflow in `wgtr.c`.


gbeeley (approved):
Looks good - thanks Lightning11wins!

PR #93, opened 2026-03-26
Fix expect.h

Lightning11wins:
There were a couple bugs with #86, mostly caused by me not knowing how to write m4 macros that _actually work._ This should fix those issues.

### DIFF
--- a/centrallix-lib/include/expect.h
+++ b/centrallix-lib/include/expect.h
@@ -11,7 +11,7 @@
 /* GNU Lesser General Public License, Version 2.1, contained in the	*/
 /* included file "COPYING".						*/
 /* 									*/
-/* Module:	Expect Branch Optimization Module (mtask.c, mtask.h)	*/
+/* Module:	Expect Branch Optimization Module (expect.h)		*/
 /* Author:	Israel Fuller						*/
 /* Date:	February 27th, 2026					*/
 /*									*/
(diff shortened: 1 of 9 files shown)


Lightning11wins:
@greptileai The previous fix was badly designed, so I've fixed the fix. Please re-review.

Lightning11wins:
@gbeeley @nboard 
PR cleared for human review.

Lightning11wins:
I realized today that guaranteeing that the return values is normalized is useful functionality, so I'm adding a TODO to make that change before we merge this.

Lightning11wins:
Alright, the functions now always normalize values.

Once again, **this PR cleared for human review.**

centrallix-lib/include/expect.h
```diff
@@ -23,19 +23,26 @@
...
-/*** Fallback: Define pass through functions to support compilers that don't
- *** have this feature.
- ***/
-#define LIKELY(x)   (x)
-#define UNLIKELY(x) (x)
+ /*** Fallback: Define pass through functions to support compilers that don't
+  *** have this feature.
+  ***/
+ #define LIKELY(x)   !!(x)
+ #define UNLIKELY(x) !!(x)
```
gbeeley: Ideally these should be (!!(x))

gbeeley (changes requested):
One small change needed - thanks!

centrallix-lib/include/expect.h
Lightning11wins: Oops, you're right. Fixed.

PR #80, opened 2025-11-25
Expand Centrallix-lib Tests

Lightning11wins:
Add additional testing for the `newmalloc`, `util`, `xarray`, `xstring`, and `clusters` libraries.
Add `test_utils.h` to make tests cleaner to read and write.
Fix a couple of minor bugs in libraries.

PR #77, opened 2025-11-14
Update Duplicate Detection

Lightning11wins:
The duplicate detection project is ready to review, although (best case), there are still a couple of things blocking it from being ready to merge.

I would appreciate a full review of all changes, as there's quite a lot here. That said, some areas may require additional special attention, so I've compiled a list of all 28 `TODO: Greg` comments below. (Note: Some of my todos assume the reader understands various pieces of nearby context / has generally read the indicated source code.)

- 3 `TODO: Greg`s in `objdrv_cluster.c`
- 1 `TODO: Greg` in `mtsession.md`.
- 1 `TODO: Greg` in `xarray.md`.
- 1 `TODO: Greg` in `xstring.md`.

Please let me know if you have any questions, comments, or concerns about my changes and design choices.

PR #97, opened 2026-04-13
Qprintf Rewrite

Lightning11wins:
`qprintf.c` had some code that confused me a lot.  For example, there was a switch statement inside an if statement inside another switch statement inside a loop inside another if statement inside the else block of different if statement inside another if statement inside another loop... causing my brain to throw a stack overflow exception.

After understanding the code (which will be useful for the [apos_autoscale](https://github.com/LightSys/centrallix/pull/90) proejct), I did some refactoring, and now it's much easier to read, in my opinion.  Also, I added some tools for improved error handling, as usual.

_Other changes_ (not in this PR):
I also wrote additional tests for qprintf, which passed, but they require several other PRs as dependencies, so I'll add them to the [centrallix-lib-tests](https://github.com/LightSys/centrallix/pull/80) branch once I update it after the [dups](https://github.com/LightSys/centrallix/pull/77) PR is merged.

### DIFF
--- a/centrallix-lib/src/qprintf.c
+++ b/centrallix-lib/src/qprintf.c
@@ -246,18 +246,20 @@ qpf_internal_FindStr(const char* haystack, size_t haystacklen, const char* needl
 int
 qpf_internal_SetupTable(pQPConvTable table)
     {
-    int i;
-    int mx = 1;
-    int n;
+    size_t mx = 1;
 
-	for(i=0;i<QPF_MATRIX_SIZE;i++)
+	for (size_t i = 0lu; i < QPF_MATRIX_SIZE; i++)
 	    {
-	    if (table->Matrix[i])
-		{
-		n = strlen(table->Matrix[i]);
-		if (n > mx) mx = n;
-		table->MatrixLen[i] = n;
-		}
+	    /*** Skip characters that don't map to different characters.
+	     *** LIKELY because most translation tables do not include
+	     *** most characters.
+	     ***/
+	    if (LIKELY(table->Matrix[i] == NULL)) continue;
+	    
+	    /** Compute the length of the string that this char maps to. **/
+	    const size_t n = strlen(table->Matrix[i]);
+	    if (n > mx) mx = n;
+	    table->MatrixLen[i] = n;
 	    }
 	table->MaxExpand = mx;
 
@@ -479,11 +481,10 @@ qpfOpenSession(void)
     {
     pQPSession s = NULL;
 
-	if (UNLIKELY(!QPF.is_init)) 
-	    {
-	    if (UNLIKELY(qpfInitialize() < 0)) return NULL;
-	    }
+	/** Ensure initialization. **/
+	if (UNLIKELY(!QPF.is_init) && UNLIKELY(qpfInitialize() < 0)) return NULL;
 
+	/** Allocate and initialize the new session. **/
 	s = (pQPSession)nmMalloc(sizeof(QPSession));
 	if (UNLIKELY(!s)) return NULL;
 	s->Errors = 0;
@@ -908,97 +909,94 @@ qpfPrintf_va(pQPSession s, char* str, size_t size, const char* format, va_list a
  *** `1`, to leave room for a null terminator, or `2` to leave room for a
  *** closing quote mark followed by a null terminator.
  *** @returns The number of chars placed in dstbuf (or the number that would
- *** have been placed if there was enough room.  Note: Does NOT return the
- *** number of chars pulled from the srcbuf!!!
+ *** have been placed if there was enough room), or -1 if an error occurs.
+ *** Note: Does NOT return the number of chars pulled from the srcbuf!!!
  ***/
 static inline int
-qpf_internal_Translate(pQPSession s, const char* srcbuf, size_t srcsize, char** dstbuf, size_t* dstoffs, size_t* dstsize, size_t limit, pQPConvTable table, qpf_grow_fn_t grow_fn, void* grow_arg, size_t min_room)
-    {
-    int rval = 0;
-    unsigned int tlen;
-    int i;
-    char* trans;
+qpf_internal_Translate(
+    pQPSession s,
+    const char* srcbuf,
+    size_t srcsize,
+    char** dstbuf,
+    size_t* dstoffs,
+    size_t* dstsize,
+    size_t limit,
+    pQPConvTable table,
+    qpf_grow_fn_t grow_fn,
+    void* grow_arg,
+    size_t min_room
+)   {
+    int n_chars_written = 0;
     int nogrow = (grow_fn == NULL);
 
-	if (UNLIKELY(srcsize >= SIZE_MAX/2/table->MaxExpand))
+	/** Check for sources that are FAR too large for us to handle. **/
+	if (UNLIKELY(srcsize >= (SIZE_MAX / 2) / table->MaxExpand))
 	    return -1;
 
-	if (LIKELY(srcsize))
+	if (LIKELY(srcsize != 0))
 	    {
-	    rval += srcsize;
-	    if ((srcsize*table->MaxExpand) <= limit && (srcsize*table->MaxExpand + min_room) <= (*dstsize - *dstoffs))
+	    n_chars_written += srcsize;
+	    const size_t max_chars_to_write = srcsize * table->MaxExpand;
+	    const size_t max_chars_needed = max_chars_to_write + min_room;
+	    const size_t chars_available = *dstsize - *dstoffs;
+	    if (max_chars_to_write <= limit && max_chars_needed <= chars_available)
 		{
 		/** Easy route - definitely enough space! **/
-		for(i=0;i<srcsize;i++)
+		for (size_t i = 0; i < srcsize; i++)
 		    {
-		    if (UNLIKELY(((trans = table->Matrix[(unsigned char)(srcbuf[i])]) != NULL)))
-			{
-			tlen = table->MatrixLen[(unsigned char)(srcbuf[i])];
-			while(*trans) (*dstbuf)[(*dstoffs)++] = *(trans++);
-			rval += (tlen-1);
-			}
-		    else
+		    const unsigned char c = (unsigned char)(srcbuf[i]);
+		    const char* translated_chars = table->Matrix[c];
+		    
+		    /*** Check if the translation table specifies to do nothing to this character.
+		     *** LIKELY because most translation tables do not include most characters.
+		     ***/
+		    if (LIKELY(translated_chars == NULL))
 			{
 			(*dstbuf)[(*dstoffs)++] = srcbuf[i];
+			continue;
 			}
+		    
+		    /** Write the translated characters to the destination buffer. **/
+		    const unsigned int translated_length = table->MatrixLen[c];
+		    while(*translated_chars) (*dstbuf)[(*dstoffs)++] = *(translated_chars++);
+		    n_chars_written += (translated_length-1);
 		    }
 		}
 	    else
 		{
 		/** Hard route - may or may not be enough space! **/
-		for(i=0;i<srcsize;i++)
+		for (size_t i = 0lu; i < srcsize; i++)
 		    {
-		    if (UNLIKELY(((trans = table->Matrix[(unsigned char)(srcbuf[i])]) != NULL)))
+		    const unsigned char c = (unsigned char)(srcbuf[i]);
+		    const char* translated_chars = table->Matrix[c];
+		    const unsigned int translated_length = (LIKELY(translated_chars == NULL)) ? 1 : table->MatrixLen[c];
+		    
+		    if (UNLIKELY(translated_length > limit))
 			{
-			tlen = table->MatrixLen[(unsigned char)(srcbuf[i])];
-			if (LIKELY(limit >= tlen))
-			    {
-			    rval += (tlen-1);
-			    if (LIKELY(!nogrow) && (LIKELY((*dstoffs)+tlen+min_room <= (*dstsize)) || 
-				  (grow_fn(dstbuf, dstsize, *dstoffs, grow_arg, (*dstoffs)+tlen+min_room))))
-				{
-				while(*trans) (*dstbuf)[(*dstoffs)++] = *(trans++);
-				limit -= tlen;
-				}
-			    else
-				{
-				QPERR(QPF_ERR_T_BUFOVERFLOW);
-				nogrow = 1;
-				}
-			    }
-			else
-			    {
-			    QPERR(QPF_ERR_T_INSOVERFLOW);
-			    rval--;
-			    }
+			QPERR(QPF_ERR_T_INSOVERFLOW);
+			n_chars_written--;
+			continue;
 			}
-		    else
+		    
+		    /** Check the available space in the destination buffer. **/
+		    n_chars_written += (translated_length-1);
+		    const size_t chars_needed = *dstoffs + translated_length + min_room;
+		    if (nogrow || (chars_needed > *dstsize && !grow_fn(dstbuf, dstsize, *dstoffs, grow_arg, chars_needed)))
 			{
-			if (LIKELY(limit > 0))
-			    {
-			    if (LIKELY(!nogrow) && (LIKELY((*dstoffs)+1+min_room <= (*dstsize)) || 
-				  (grow_fn(dstbuf, dstsize, *dstoffs, grow_arg, (*dstoffs)+1+min_room))))
-				{
-				(*dstbuf)[(*dstoffs)++] = srcbuf[i];
-				limit--;
-				}
-			    else
-				{
-				QPERR(QPF_ERR_T_BUFOVERFLOW);
-				nogrow = 1;
-				}
-			    }
-			else
-			    {
-			    QPERR(QPF_ERR_T_INSOVERFLOW);
-			    rval--;
-			    }
+			QPERR(QPF_ERR_T_BUFOVERFLOW);
+			nogrow = 1;
+			continue;
 			}
+		    
+		    /** Write the translated characters to the destination buffer. **/
+		    if (LIKELY(translated_chars == NULL)) (*dstbuf)[(*dstoffs)++] = srcbuf[i];
+		    else while(*translated_chars) (*dstbuf)[(*dstoffs)++] = *(translated_chars++);
+		    limit -= translated_length;
 		    }
 		}
 	    }
 
-    return rval;
+    return n_chars_written;
     }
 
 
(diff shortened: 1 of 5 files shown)


Lightning11wins:
@greptileAI Please review this PR.

centrallix-lib/src/qprintf.c
```diff
@@ -767,681 +967,740 @@ qpfPrintf_va(pQPSession s, char* str, size_t size, const char* format, va_list a
...
+			const int is_only_followed_by_nlen = (n_specs - i == 2 || specchain[i + 1] == QPF_SPEC_T_NLEN);
+			if (!is_final_spec && !is_only_followed_by_nlen)
+			    {
+			    QPERR(QPF_ERR_T_NOTIMPL);
+			    rval = -ENOSYS;
+			    goto error;
+			    }
+			
+			/** Determine the max amount of characters that we can write. **/
+			size_t maxdst = (is_only_followed_by_nlen) ? specchain_n[i + 1] : INT_MAX;
```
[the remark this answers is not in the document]
Lightning11wins: Good catch, fixed.

centrallix-lib/src/qprintf.c
```diff
@@ -767,681 +967,740 @@ qpfPrintf_va(pQPSession s, char* str, size_t size, const char* format, va_list a
...
+	if (UNLIKELY(!QPF.is_init) && UNLIKELY(qpfInitialize() < 0)) return -ENOMEM;
+	
+	/** Use a null session (that does nothing with errors) if no session is provided. **/
+	QPSession null_session;
+	if (s == NULL)
 	    {
-	    if (qpfInitialize() < 0) return -ENOMEM;
+	    null_session.Errors = 0;
+	    s = &null_session;
 	    }
```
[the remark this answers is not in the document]
Lightning11wins: Oops, fixed.

centrallix-lib/src/qprintf.c
```diff
@@ -767,681 +967,740 @@ qpfPrintf_va(pQPSession s, char* str, size_t size, const char* format, va_list a
...
+			/** Add closing quote (if requested). **/
+			if (quote)
+			    {
+			    const size_t space_needed = dest_offset + 2lu;
+			    if (space_needed > *dest_size && !grow_fn(dest, dest_size, dest_offset, grow_arg, space_needed))
+				{
+				QPERR(QPF_ERR_T_BUFOVERFLOW);
+				no_grow = true;
+				}
+			    else (*dest)[dest_offset++] = quote;
```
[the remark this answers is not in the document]
Lightning11wins: Good catch, fixed.

centrallix-lib/src/qprintf.c
```diff
@@ -166,31 +186,78 @@ typedef struct
...
+    int shift = 0;
+    
+	while ((n & 1) == 0) 
+	    {
+	    n >>= 1;
+	    shift++;
+	    }
+    
+    return shift;
+    }
```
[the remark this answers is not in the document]
Lightning11wins: Sure, done.

Lightning11wins:
@greptileai I think I resolved your comments. Please rereview. Are there any other concerns you have?

centrallix-lib/src/qprintf.c
```diff
@@ -767,681 +969,742 @@ qpfPrintf_va(pQPSession s, char* str, size_t size, const char* format, va_list a
...
+	    copied += copy_len;
+	    const size_t space_needed = dest_offset + copy_len + 1lu;
+	    if (UNLIKELY(space_needed > SIZE_MAX/2))
+		{
+		QPERR(QPF_ERR_T_BUFOVERFLOW);
+		rval = -EINVAL;
+		goto error;
+		}
+	    if (UNLIKELY(no_grow)) copy_len = 0;
+	    if (UNLIKELY(space_needed > *dest_size && !grow_fn(dest, dest_size, dest_offset, grow_arg, space_needed)))
```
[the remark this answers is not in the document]
Lightning11wins: Fixed.

Lightning11wins:
@greptileai I think I resolved your comments. Please rereview. Are there any other concerns you have?

centrallix-lib/src/qprintf.c
```diff
@@ -767,681 +969,742 @@ qpfPrintf_va(pQPSession s, char* str, size_t size, const char* format, va_list a
...
+			    if (UNLIKELY(no_grow || (space_needed > *dest_size && !grow_fn(dest, dest_size, dest_offset, grow_arg, space_needed))))
+				{
+				QPERR(QPF_ERR_T_BUFOVERFLOW);
+				no_grow = true;
+				}
+			
+			    /** Write the closing quote, even if a buffer overflow occurred. **/
+			    /** We passed min_room = 2, so there will be enough space for it. **/
+			    (*dest)[dest_offset++] = quote;
+			    copied++;
```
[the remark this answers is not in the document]
Lightning11wins: @greptileai
Interesting, but making that change results in `test_qprintf_51` and `test_qprintf_52` failing because they ensure that the closing quote is _always_ included, even if an overflow occurs while adding the quoted text. This is required functionality.

Can you propose an alternate solution that maintains this required functionality while fixing the perceived issue?

Lightning11wins:
@greptileai I've made the changes you've requested but I haven't tested or pushed my code yet. Could you write a test case, following the same style established in `test_qprintf_51.c` that will ensure that will verify that this issue is fixed? Thus, it should fail on the previous implementation (the one that did not move the pointer AND the one that did not check `no_grow`), but it should pass on the current implementation that you showed me, which handles the situation correctly.

Lightning11wins:
@greptileai I think I resolved your comments, and I also added your test case with some clean up, bug fixing (on the return value), and other improvements. Please rereview. Are there any other concerns you have?

centrallix-lib/src/qprintf.c
```diff
@@ -767,681 +969,748 @@ qpfPrintf_va(pQPSession s, char* str, size_t size, const char* format, va_list a
...
+	    /** If this specifier is ignored, we're done. Skip all filtering/writing logic. **/
+	    if (UNLIKELY(ignore)) continue;
+	    
+	    /** Check for invalid length. **/
+	    if (UNLIKELY(copy_len < 0))
+		{
+		/** This error case appears to be unreachable. **/
+		rval = -EINVAL;
+		QPERR(QPF_ERR_T_BADLENGTH);
+		goto error;
```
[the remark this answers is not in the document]
Lightning11wins: We don't need to error check the built-in C `print()` functions in this project. Please remember that in the future.

Lightning11wins:
@greptileai I think I resolved your comments. Please rereview. Are there any other concerns you have?

centrallix-lib/src/qprintf.c
```diff
@@ -767,681 +970,739 @@ qpfPrintf_va(pQPSession s, char* str, size_t size, const char* format, va_list a
...
+			     
+			    /*** Write the closing quote with space for the
+			     *** null-terminator, even if a buffer overflow
+			     *** has occurred, by moving the offset back and
+			     *** overwriting some of the end of the provided
+			     *** string, if needed.
+			     ***/
+			    if (space_needed > *dest_size)
+				dest_offset = *dest_size - 2;
+			    (*dest)[dest_offset++] = quote;
```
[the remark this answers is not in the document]
Lightning11wins: Good catch, I've added a test case for this.

Lightning11wins:
@greptileai I think I resolved your comments, and I also added your test case with some clean up, bug fixing (on the return value), and other improvements. Please rereview. Are there any other concerns you have?

Lightning11wins:
@greptileai I just put some finishing touches on the project. Could you review this please?

Lightning11wins:
**This PR is cleared for human review.**

centrallix-lib/src/qprintf.c
```diff
@@ -748,700 +1001,809 @@ qpfPrintf(pQPSession s, char* str, size_t size, const char* format, ...)
...
- *** null terminator, for instance.
+ *** @param s The qprintf session in use.
+ *** @param srcbuf The source string representation to translate and copy.
+ *** @param srcsize The length of `srcbuf` in bytes.
+ *** @param dstbuf A pointer to a location where the resulting string pointer
+ *** 	should be saved.
+ *** @param dstoffs The number of bytes to skip at the start of `dstbuf`
+ *** 	before writing the result.
+ *** @param dstsize The size of the currently allocated string at `dstbuf`.
+ *** @param limit The maximum amount that the destination string can grow past
```
gbeeley: `limit` is a hard limit on the maximum size of the insertion.  This is independent of the ability to grow the destination: if nLEN / *LEN is used in the spec chain, that places a hard limit on the amount of data that can be inserted for the one specific conversion/translation.  The destination may be able to be grown larger (or may otherwise have plenty of space already), but that's a separate issue.

Besides fixing this function doc, review the refactored code to ensure it is consistent with this use of `limit`.

centrallix-lib/src/qprintf.c
```diff
@@ -748,700 +1001,809 @@ qpfPrintf(pQPSession s, char* str, size_t size, const char* format, ...)
...
-			if (*format >= '0' && *format <= '9')
+			/** need at least one spec **/
+			rval = -EINVAL;
+			QPERR(QPF_ERR_T_BADFORMAT);
+			goto error;
+			}
+		    
+		    /** Invalid spec: Skip to printing. **/
+		    format--;
+		    break;
```
gbeeley, line 1467: This appears to ignore filters that aren't found / aren't handled.  This at least deserves a warning message (yes, this was from the original code, not from this PR). However, the commit history indicates this was a deliberate choice due to `&nbsp;` occurring in the format string.  It seems that only happens in a few places and could easily be worked around via `%&`.  Not sure if there are other implications.

centrallix-lib/src/qprintf.c
```diff
@@ -748,700 +1001,809 @@ qpfPrintf(pQPSession s, char* str, size_t size, const char* format, ...)
...
+		    }
+		
+		/** Later items in the specifier chain must be filter specifiers. **/
+		startspec = QPF_SPEC_T_STARTFILT;
+		endspec = QPF_SPEC_T_ENDFILT;
+		}
+	    while (format[0] == '&' && format++); /* Loop as long as there are '&' chars to consume. */
```
gbeeley: Is there a reason for allowing consecutive `&` characters without filters?

centrallix-lib/src/qprintf.c
```diff
@@ -748,700 +1001,809 @@ qpfPrintf(pQPSession s, char* str, size_t size, const char* format, ...)
...
+	    
+	    /** Get the data using the source spec. **/
+	    char tmp_buf[318]; /* 318 characters are needed to print DBL_MAX. */
```
gbeeley, line 1477: This should be a #define somewhere.

centrallix-lib/src/qprintf.c
```diff
@@ -748,700 +1001,809 @@ qpfPrintf(pQPSession s, char* str, size_t size, const char* format, ...)
...
+			}
+		    copy_len = qpf_internal_itoa(tmp_buf, sizeof(tmp_buf), int_val);
+		    strval = tmp_buf;
+		    break;
+		    }
+		
+		case QPF_SPEC_T_LL:
+		    {
+		    const long long ll_val = va_arg(ap, long long);
+		    copy_len = (size_t)snprintf(tmp_buf, sizeof(tmp_buf), "%lld", ll_val);
```
gbeeley: Hopefully snprintf() doesn't fail here.  Good to check.  (similar in other nearby lines)

centrallix-lib/src/qprintf.c
```diff
@@ -748,700 +1001,809 @@ qpfPrintf(pQPSession s, char* str, size_t size, const char* format, ...)
...
+			    goto error;
+			    }
+			copied += num_bytes;
+			copy_len = 0;
+			break;
+			}
+		    
+		    case QPF_SPEC_T_QUOT:    table = &QPF.quote_matrix;    min_room = 2; quote = '\''; goto use_table;
+		    case QPF_SPEC_T_DQUOT:   table = &QPF.quote_matrix;    min_room = 2; quote = '"';  goto use_table;
+		    case QPF_SPEC_T_ESCQ:    table = &QPF.quote_matrix;    goto use_table;
```
gbeeley: Let's avoid `goto` in this usage (not an error/exit/retry situation, not a state machine).

centrallix-lib/src/qprintf.c
```diff
@@ -748,700 +1001,809 @@ qpfPrintf(pQPSession s, char* str, size_t size, const char* format, ...)
...
+		    }
+		}
+	    
+	    /** If this specifier is ignored, we're done. Skip all filtering/writing logic. **/
+	    if (UNLIKELY(ignore)) continue;
+	    
+	    /** Handle filters. **/
+	    pQPConvTable table;
+	    size_t min_room = 1;
+	    char quote = 0;
```
gbeeley: Setting `min_room` and `quote` here works in the current implementation where we don't allow the chaining of translation/filter specs yet.  But they probably should be moved into the spec loop so this doesn't become a latent bug if chaining of multiple filters is implemented in the future.

centrallix-lib/src/qprintf.c
```diff
@@ -748,700 +1001,809 @@ qpfPrintf(pQPSession s, char* str, size_t size, const char* format, ...)
...
 				}
-			    memcpy((*str)+cpoffset, strval, cplen);
+			    else (*dest)[dest_offset++] = quote;
+			    copied++;
 			    }
-
-			/** Update string counters **/
-			cpoffset += cplen;
+			
+			/** Translate the string content using the table selected above. **/
```
gbeeley: There used to be a check here that verified that n_chars returned from Translate matched the change in the dest_offset.  The check appears to be aimed at preventing multiple successive failed grow function calls if a grow already failed, as in other areas of the code.

centrallix-lib/src/qprintf.c
```diff
@@ -748,700 +1001,809 @@ qpfPrintf(pQPSession s, char* str, size_t size, const char* format, ...)
...
+	    copied += copy_len;
+	    const size_t space_needed = dest_offset + copy_len + 1lu;
+	    if (UNLIKELY(space_needed > SIZE_MAX/2))
+		{
+		QPERR(QPF_ERR_T_BUFOVERFLOW);
+		rval = -EINVAL;
+		goto error;
+		}
+	    if (UNLIKELY(no_grow)) copy_len = 0;
+	    if (UNLIKELY(no_grow || (space_needed > *dest_size && !grow_fn(dest, dest_size, dest_offset, grow_arg, space_needed))))
```
gbeeley: This space needed / error logic appears to be different from the original and different from the other plain string copying code early in this function.  Do a double check review for correctness on this.  The switch from a computed-each-time space needed expression to a computed-once space needed expression could create some divergence in the logic from the original.

centrallix-lib/src/qprintf.c
Lightning11wins: I was a little confused by this but now understand much better and I've refined the doc comments.

Lightning11wins: I've just conducted a review and I believe my refactored code implements this feature correctly.

centrallix-lib/src/qprintf.c
Lightning11wins: I don't think I do? An invalid specifier breaks out of the `do {} while()` loop (see above). Thus, additional `&`s after a specifier are not consumed. I think this is the same behavior as `master`.

For example, the format specifier `x%STR&&y&` (with `"a<b"` passed) results in: `xa<b&&y&`.

centrallix-lib/src/qprintf.c
Lightning11wins: After looking into this, it looks like you were right. `&nbsp` _directly_ after a specifier will trigger the warning but `%&nbsp` does not.

According to Claude, this contract change only affects two places in production (both in `htdrv_tab.c`). I've updated those on this branch so that the merge conflict on `apos_autoscale9-slim5` will remind me to scan it for this issue, although I don't expect to find much.

Lightning11wins: An unexpected side effect is that this warning caused the tests to fail by lockup because printing this warning hundreds of thousands of times slows down one of the tests a lot.

This doesn't seem like a real bug detected from the test (calling `qprintf()` with a format that triggers this warning hundreds of thousands of times per second doesn't seem like a usecase we need to support), so I changed the test to pipe stderr to `/dev/null`, fixing the lockup.

centrallix-lib/src/qprintf.c
Lightning11wins: I couldn't find a pre-existing define for this so I defined it above. We should probably have a lint rule against magic numbers in code.

Lightning11wins: I added a rule against magic numbers to the style guide.

centrallix-lib/src/qprintf.c
Lightning11wins: On `master`, these values are defined in the function's top-level scope. I didn't understand them super well so I moved them to the lowest scope I felt I safely could. Thanks for the clarification, I've fixed this.

centrallix-lib/src/qprintf.c
Lightning11wins: Oh, great. This refactor is about to wreck the git blame.

centrallix-lib/src/qprintf.c
Lightning11wins: Added error handling.

centrallix-lib/src/qprintf.c
Lightning11wins: I've refactored to use an if statement instead and I think the resulting code is better, thanks for the feedback.

centrallix-lib/src/qprintf.c
Lightning11wins: Ahh, looks like this regression came out of an incorrect fix I made in the past that weakened this check. Fixed.

centrallix-lib/src/qprintf.c
Lightning11wins: I don't fully understand this logic, but I think I fixed it.

Lightning11wins:
We should possibly consider making the repeated grow logic a helper function. Despite my best effort to make it easy to read, it's one of the most confusing parts of the refactored code.

Lightning11wins:
@gbeeley I believe I've addressed all the concerns from your review.

Lightning11wins:
I've fixed several other bugs that I found on my own, now, too. I think that's everything, though.

Lightning11wins:
Added `%ULL` and `%BOOL`.

centrallix-lib/src/qprintf.c
```diff
@@ -513,7 +706,7 @@ qpf_internal_base64encode(pQPSession s, const char* src, size_t src_size, char**
     const unsigned char* srcptr = (const unsigned char*)src;
     const unsigned char* origsrc = (const unsigned char*)src;
     char* dstptr;
-    int req_size = ((src_size+2) / 3) * 4 + *dst_offset;
+    int req_size = ((src_size+2) / 3) * 4 + *dst_offset + 1; /** +1 leaves room for the null-terminator. **/
```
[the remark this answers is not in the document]
Lightning11wins: Fixed.

PR #98, opened 2026-04-24
Include mssError() when used

Lightning11wins:
Some files call `msError()` without including `cxlib/mtsession.h`. This latent mistake caused compiler errors when I made an otherwise safe change in code I plan to commit to for #77 once it's not being reviewed.

I'm making this a new PR because I don't want to mix even more mostly unrelated changes into the `dups` branch, and I think these changes will be obviously safe on their own, so they should be trivial to review, approve, and merge.


Lightning11wins:
**Cleared for human review.**

Lightning11wins:
**This PR is cleared for human review.**

gbeeley (approved):
Looks good - thanks!

PR #99, opened 2026-05-07
Remove Obsolete `sys_osml` Widget Driver

Lightning11wins:
According to our discussion on Riot, this widget is obsolete and was never fully implemented, so I've removed it to reduce clutter in the codebase.


gbeeley (approved):
Looks good - thanks

PR #106, opened 2026-05-18
Test_obj `exit` command

Lightning11wins:
This PR adds the `exit` command alias for the test_obj `quit` command because I kept typing the wrong command.


gbeeley (approved):
Looks good; thanks!

PR #107, opened 2026-05-20
Fix `ORDER BY` Seg Fault

Lightning11wins:
This PR fixes a segmentation fault caused by some queries that used order by.

Thank you to Greg for quickly tracking down the issue so that I could write this simple fix.

### DIFF
--- a/centrallix/multiquery/multiq_orderby.c
+++ b/centrallix/multiquery/multiq_orderby.c
@@ -269,6 +269,8 @@ mqobAnalyzeAfterGroup(pQueryStatement stmt)
 	    for(i=0;i<order_qs->Children.nItems;i++)
 		{
 		order_item = (pQueryStructure)(order_qs->Children.Items[i]);
+		if (order_item == NULL || order_item->Expr == NULL) continue;
+		
 		if (sep_groupby || (order_item->Expr && order_item->Expr->AggLevel == 1))
 		    {
 		    /** Found one.  Squirrel it away in our order-by list. **/
@@ -277,7 +279,6 @@ mqobAnalyzeAfterGroup(pQueryStatement stmt)
 			mssError(1, "MQOB", "Too many ORDER BY expressions (max %d)", MQ_MAX_ORDERBY);
 			goto error;
 			}
-		    if (order_item->Expr == NULL) continue;
 		    qe->OrderBy[n_orderby++] = exp_internal_CopyTree(order_item->Expr);
 		    }
 		}
--- a/centrallix/multiquery/multiq_orderby.c
+++ b/centrallix/multiquery/multiq_orderby.c
@@ -16,7 +16,7 @@
 /* Centrallix Application Server System 				*/
 /* Centrallix Core       						*/
 /* 									*/
-/* Copyright (C) 1999-2010 LightSys Technology Services, Inc.		*/
+/* Copyright (C) 1999-2026 LightSys Technology Services, Inc.		*/
 /* 									*/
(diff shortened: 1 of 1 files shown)


centrallix/multiquery/multiq_orderby.c
```diff
@@ -277,7 +277,8 @@ mqobAnalyzeAfterGroup(pQueryStatement stmt)
 			mssError(1, "MQOB", "Too many ORDER BY expressions (max %d)", MQ_MAX_ORDERBY);
 			goto error;
 			}
-		    qe->OrderBy[n_orderby++] = exp_internal_CopyTree(order_item->Expr);
+		    pExpression expr = order_item->Expr;
+		    qe->OrderBy[n_orderby++] = (expr != NULL) ? exp_internal_CopyTree(expr) : NULL;
```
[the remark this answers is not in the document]
Lightning11wins: That sounds reasonable, updated and fixed.

centrallix/multiquery/multiq_orderby.c
nboard: The greptile summary notes:
"The segfault fix is correct in intent, but the NULL guard is ordered after the bounds check, which can produce a spurious error for queries with exactly 24 valid ORDER BY expressions followed by a NULL-expression item when sep_groupby is true."
I think we need to decide if NULL values should count towards the total or not, because having it only count on the last iteration feels weird (better than crashing, of course). 

Consider:

```suggestion
		if (order_item->Expr && (sep_groupby ||  order_item->Expr->AggLevel == 1))
```
And then removing the line added below with the continue. 

While the current  version makes it seem like the author wanted to allow NULL values if `sep_groupby` was true, the fact it crashes forces us to do nothing in that case regardless.

nboard (dismissed):

centrallix/multiquery/multiq_orderby.c
Lightning11wins: Ahh, I see. I applied that change, but I've architected it a little differently because I thought the condition in the if statement was getting too long and this would be more clear.

Also, I threw in some code to prevent a seg fault if `order_item`, which looks like it will never happen, but we'll check it just to be safe.

Change:
```diff
@@ -269,6 +269,8 @@
 	    for(i=0;i<order_qs->Children.nItems;i++)
 		{
 		order_item = (pQueryStructure)(order_qs->Children.Items[i]);
+		if (order_item == NULL || order_item->Expr == NULL) continue;
+		
 		if (sep_groupby || (order_item->Expr && order_item->Expr->AggLevel == 1))
 		    {
 		    /** Found one.  Squirrel it away in our order-by list. **/
@@ -277,7 +279,6 @@
 			mssError(1, "MQOB", "Too many ORDER BY expressions (max %d)", MQ_MAX_ORDERBY);
 			goto error;
 			}
-		    if (order_item->Expr == NULL) continue;
 		    qe->OrderBy[n_orderby++] = exp_internal_CopyTree(order_item->Expr);
 		    }
 		}
```

Lightning11wins: Commit pushed.

nboard: `order_item->Expr` gets checked twice here - I think you can simplify that second if to simply:
```c
if (sep_groupby || order_item->Expr->AggLevel == 1)
```
Checking if order_item is NULL seems like a good addition here 👍

Lightning11wins: Oh, true. Cleaned up.

gbeeley (approved):
Looks good, thanks Lightning11wins!

PR #108, opened 2026-05-20
Update test_prt.c

tyounglightsys:
Linking was giving issues on newer compilers.  Needed to explicitly state that CxGlobals was external.


nboard (commented):
Looks good, and compiles fine on my dev vm as well

nboard (approved):

PR #109, opened 2026-05-26
Add AI Files

Lightning11wins:
Add the following files that are useful to provide context for AI agents.

Files added:
- `AGENTS.md`: General standard for providing context to most types of agents.
- `CLAUDE.md`: A symlink to `AGENTS.md`, because Claude uses its own standard instead of `AGENTS.md`.
- `.cursorrules`: A file of rules to give to the Cursor agent specifically. Currently, this is just used to patch over some Cursor UI bugs.

**Note**: These files should be minimal since they're included _every_ time you start a new thread with an agent. Thus, we need to avoid including TONS of context that agents won't need in most cases, or that they can easily find out on their own.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -12,3 +12,10 @@ While it may be very useful to read the following generated files, do not edit t
 - Centrallix-lib/Makefile
 - Centrallix/configure(.sh)
 - Centrallix-lib/configure(.sh)
+
+# Terms
+- **Centrallix** - An open-source web application server and data management engine developed by LightSys. It provides data abstraction, structural embedding, and dynamic widget-based HTML generation (DHTML), serving as the platform on which apps like Kardia are built.
+- **Kardia** - A web-based nonprofit ministry management application built on the Centrallix platform. It provides donor tracking, financial management, and ministry operations for Christian mission organizations.
+- **Object System (ObjectSystem)** - Centrallix's high-level filesystem analog: a unified tree of typed objects allowing controlled read and write access to files, database rows, directories, APIs, reports, and more, each accessible through a consistent path-based API (the OSML). ObjectSystem Drivers (OSDs) provide the intelligence to interpret different object types, allowing disparate data sources to appear as a single coherent hierarchy.
+- **Widget** - A UI component in Centrallix's DHTML application layer, either visual (e.g. buttons, edit boxes, tables) or nonvisual (e.g. forms, event connectors). Widgets are defined as objects in the ObjectSystem and rendered to HTML/JavaScript by widget drivers on the server side. (For info on using specific widgets, see `centrallix-sysdoc/widgets.md`)
+For other terminology or to find more detailed info, read `centrallix-sysdoc/Terminology.md`.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -8,10 +8,10 @@ When making meaningful changes to a file (more than a few lines), update the cop
 
 ## Generated files
 While it may be very useful to read the following generated files, do not edit them since changes will be overwritten by builds.
-- Centrallix/Makefile
-- Centrallix-lib/Makefile
-- Centrallix/configure(.sh)
-- Centrallix-lib/configure(.sh)
+- centrallix/Makefile
+- centrallix-lib/Makefile
+- centrallix/configure
+- centrallix-lib/configure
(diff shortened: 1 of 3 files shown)


Lightning11wins:
I added a bit more detail to help agents do tasks like creating or editing app files. I'm trying to keep everything as concise as possible by telling agents to reference existing docs. I'd prefer not to pollute context with info if it might not be relevant in some cases.

AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Generated files
+While it may be very useful to read the following generated files, do not edit them since changes will be overwritten by builds.
+- centrallix/Makefile
+- centrallix-lib/Makefile
+- centrallix/configure
+- centrallix-lib/configure
+
+## Searching
+Instead of searching with `grep`, use `rg` (if available). It's faster, but remember that it ignores `.gitignored` files, hidden files, binaries, etc. This project includes symlinks (both to files and directories), so ensure they are traversed when searching (aka. with `rg -L` or alternatives). Many centrallix structure files (e.g. `.app`, `.cmp`, etc.) are treated as binary files by `rg`, add `-a` when searching them. You may need to combine both flags to navigate these challenges, e.g. `rg -La "pattern"`, or fall back to `grep` if you encounter issues.
```
nboard, line 17: It looks like our vms do not come with `rg` by default. Do agents get enough of an advantage from it that we should consider changing it? Or should we switch to just recommending Grep rather than having the agents fail to use ripgrep?

AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+# Terms
+
+- **Centrallix** - An open-source web application server and data management engine developed by LightSys. It provides data abstraction, structural embedding, and dynamic widget-based HTML generation (DHTML), serving as the platform on which apps like Kardia are built.
+
+- **Kardia** - A web-based nonprofit ministry management application built on the Centrallix platform. It provides donor tracking, financial management, and ministry operations for Christian mission organizations.
```
nboard: Could we potentially trim this down to just say its an app built on centrallix? If we're trying to keep the context minimal, the purpose of an app outside the centrallix project feels extraneous. 

Unless you were intending to have this act as the sole agents file for both repos?

AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+
+
+# Building Apps
+
+## Where to look
+- **Widget properties, children, events, actions**: `centrallix-doc/Widgets/widgets.xml` covers all widgets.
+- **Structure file syntax** (.app, .cmp, and others): `centrallix-doc/StructureFile.txt`.
+- **Forms** (form modes, element callbacks): `centrallix-sysdoc/HTFormsInterface.md`.
+- **Widget tree / Wgtr module** (server-side tree building, verification, component loading): `centrallix-sysdoc/WidgetTree.md`.
+- **Practical examples**: `centrallix-os/samples/*.app`.
```
nboard, line 40: Would including `report.xml` be a good idea as well?

nboard (commented):
Looks good overall! Thanks for putting this together. I have a few ideas to consider, but no required changes.

AGENTS.md
Lightning11wins: It looks like Claude Code bundles it (so the agent has access), and the new Rocky 9 VMs will come with it installed, too.

AGENTS.md
Lightning11wins: I've used Claude Code to build apps in Kardia through the symlink, but you're right that a separate `agents.md` might be good for that because we could tailor it more. For now, though, I'd like to keep supporting that workflow with this file.

Lightning11wins: That said, you're right that we could trim this a lot because the AI likely rarely needs all that context about what "Kardia" is. Updated.

Lightning11wins: A few of these sections could be trimmed a bit, let me do that.

AGENTS.md
Lightning11wins: Forgot about that, added.

AGENTS.md
nboard: Looks good!

gbeeley (approved):
Looks good - thanks Lightning11wins and nboard!

PR #110, opened 2026-05-26
Fix: Add Secure flag to CXLH login hash cookie

Lightning11wins:
The CXLH cookie (used by `NHT_AUTH_HTTPSTRICT`) was missing the Secure attribute. This allows the browser to transmit the cookie over an unencrypted connection on HTTPS deployments that also accept HTTP.

This fix mirrors the existing pattern established for the session cookie pattern: Secure is appended to the CXLH Set-Cookie header when `conn->UsingTLS` is true, and omitted on plain HTTP connections.


gbeeley (approved):
This isn't an auth vulnerability, but the patch looks fine - thanks!

PR #112, opened 2026-06-01
Replace `cxjs_eval()` by Adding Other Missing `cxjs_*()` Functions

Lightning11wins:
This PR removes `cxjs_eval()` because it was unsafe and adds the following missing `cxjs_*()` functions to help fill in the resulting functionality gap:
- `cxjs_replicate()`
- `cxjs_abs()`
- `cxjs_round()`
- `cxjs_truncate()`
- `cxjs_constrain()`
- `cxjs_rand()`
- `cxjs_sqrt()`
- `cxjs_square()`
- `cxjs_power()`
- `cxjs_degrees()`
- `cxjs_radians()`


centrallix-os/sys/js/ht_render.js
```diff
@@ -310,14 +310,6 @@ function cxjs_right(s,l)
     if (s == null || l == null) return null;
     return s.substr(s.length-l);
     }
-function cxjs_eval(x)
```
gbeeley, line 313: Let's re-add cxjs_eval() and have it only process a few basic things, and do those consistently with how eval() in CXSQL works:

- a number (integer or floating point)
- a quoted string, single or double quotes
- :property, only if there is an obvious current object context (like event params in a connector), and only if the second param is not provided or is null or contains an uppercase "C" (allow current object references).
- :object:property, only if the second param is not provided or is null or contains an uppercase "O" (allow object references).

With all of these, parse with care. :)

If you want to do this as a separate PR, please comment so here and I'll proceed with this PR.

gbeeley (changes requested):
One note to resolve - thanks!

centrallix-os/sys/js/ht_render.js
Lightning11wins: `eval()` sounds like it will be more complex than the relatively basic functions in this PR. Let's make it a separate PR.

Lightning11wins: Alright, I wrote a safe implementation of `eval()` in #116.

gbeeley (approved):
Looks good - thanks lightning11wins!

PR #116, opened 2026-06-05
Replace `cxjs_eval()` with a Safe Implementation

Lightning11wins:
This PR will add a safe client-side implementation of eval, replacing the previous unsafe version.

### DIFF
--- a/centrallix-os/sys/js/ht_render.js
+++ b/centrallix-os/sys/js/ht_render.js
@@ -313,13 +313,18 @@ function cxjs_right(s,l)
 
 function cxjs_eval(_context, _this, expr, permflags)
     {
-    console.log('eval', _context, _this, expr, permflags);
     if (expr === null || expr === undefined) return null;
-    expr_str = String(expr).trim();
+    const expr_str = String(expr).trim();
     if (expr_str === '') return null;
     
     const first_char = expr_str.charAt(0);
     const last_char = expr_str.charAt(expr_str.length - 1);
+    
+    // Eval boolean and null literals.
+    const expr_lower = expr_str.toLowerCase();
+    if (expr_lower === 'true') return true;
+    if (expr_lower === 'false') return false;
+    if (expr_lower === 'null') return null;
     
     // Eval integer.
     if (/^-?\d+$/.test(expr_str))
@@ -365,6 +370,7 @@ function cxjs_eval(_context, _this, expr, permflags)
 	}
     
     console.warn("Failed to eval expression: \"" + expr_str + "\"");
+    
     return null;
     }
 
--- a/centrallix-os/sys/js/ht_render.js
+++ b/centrallix-os/sys/js/ht_render.js
@@ -310,14 +310,64 @@ function cxjs_right(s,l)
     if (s == null || l == null) return null;
     return s.substr(s.length-l);
     }
-function cxjs_eval(x)
+
+function cxjs_eval(_context, _this, expr, permflags)
     {
-    var _this = null;
-    var _context = null;
-    if (x == null) return null;
-    if (typeof x == 'object') x = x.toString();
-    return eval(x);
+    console.log('eval', _context, _this, expr, permflags);
+    if (expr === null || expr === undefined) return null;
+    expr_str = String(expr).trim();
+    if (expr_str === '') return null;
+    
+    const first_char = expr_str.charAt(0);
+    const last_char = expr_str.charAt(expr_str.length - 1);
+    
+    // Eval integer.
+    if (/^-?\d+$/.test(expr_str))
+	return parseInt(expr_str, 10);
+    
+    // Eval floating point.
+    if (/^-?\d+\.\d*$/.test(expr_str) || /^-?\d*\.\d+$/.test(expr_str))
+	return parseFloat(expr_str);
+    
+    // Eval double-quoted string.
+    if (expr_str.length >= 2 && first_char === '"' && last_char === '"')
+	return expr_str.substring(1, expr_str.length - 1);
+    
+    // Eval single-quoted string.
+    if (expr_str.length >= 2 && first_char === "'" && last_char === "'")
+	return expr_str.substring(1, expr_str.length - 1);
+    
+    // Eval :property or :object:property
+    if (first_char === ':')
+	{
+	const rest = expr_str.substring(1);
+	const colon2 = rest.indexOf(':');
+	if (colon2 < 0)
+	    {
+	    // :property - current object (event params) reference
+	    if (_this != null && (!permflags || permflags.indexOf('C') >= 0))
+		{
+		const v = _this[rest];
+		return (typeof v !== 'undefined') ? v : null;
+		}
+	    }
+	else
+	    {
+	    // :object:property
+	    if (_context != null && (!permflags || permflags.indexOf('O') >= 0))
+		{
+		const objname = rest.substring(0, colon2);
+		const propname = rest.substring(colon2 + 1);
+		const obj = wgtrGetNode(_context, objname);
+		if (obj) return wgtrGetProperty(obj, propname);
+		}
+	    }
+	}
+    
+    console.warn("Failed to eval expression: \"" + expr_str + "\"");
+    return null;
     }
+
 function cxjs_isnull(v,d)
     {
     if (v == null)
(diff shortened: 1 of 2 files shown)


centrallix-os/sys/js/ht_render.js
```diff
@@ -310,14 +310,64 @@ function cxjs_right(s,l)
...
-    var _this = null;
-    var _context = null;
-    if (x == null) return null;
-    if (typeof x == 'object') x = x.toString();
-    return eval(x);
+
+function cxjs_eval(_context, _this, expr, permflags)
+    {
+    console.log('eval', _context, _this, expr, permflags);
+    if (expr === null || expr === undefined) return null;
```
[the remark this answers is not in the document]
Lightning11wins: Oh wow, it was. Whoops!

centrallix-os/sys/js/ht_render.js
```diff
@@ -310,14 +310,64 @@ function cxjs_right(s,l)
...
+    if (/^-?\d+\.\d*$/.test(expr_str) || /^-?\d*\.\d+$/.test(expr_str))
+	return parseFloat(expr_str);
+    
+    // Eval double-quoted string.
+    if (expr_str.length >= 2 && first_char === '"' && last_char === '"')
+	return expr_str.substring(1, expr_str.length - 1);
+    
+    // Eval single-quoted string.
+    if (expr_str.length >= 2 && first_char === "'" && last_char === "'")
+	return expr_str.substring(1, expr_str.length - 1);
```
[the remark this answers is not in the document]
Lightning11wins: Nice catch, fixing it now.

centrallix-os/sys/js/ht_render.js
```diff
@@ -310,14 +310,70 @@ function cxjs_right(s,l)
...
     }
-function cxjs_eval(x)
-    {
-    var _this = null;
-    var _context = null;
-    if (x == null) return null;
-    if (typeof x == 'object') x = x.toString();
-    return eval(x);
+
+function cxjs_eval(_context, _this, expr, permflags)
```
gbeeley: I agree with Greptile on the 3rd and 4th arguments.  We should either flag the presence of those or handle them.  Handling them would be ideal; that would mean:
- :property - current object property reference would use the object named in parameter 3 if param 3 is supplied and not null (contingent of course on "C" being in the perm string or the perm string being absent)
- ::property - parent object property reference would use the object named in parameter 4 if param 4 is supplied and not null (contingent on "P" being in the perm string or the perm string being absent)
I could see the above being useful in certain types of applications involving data import or dynamic functionality.  But writing an entire CXSQL expression parser and compiler in JavaScript isn't on the menu yet.

gbeeley (changes requested):
Need to address 3rd/4th param consistency with server per Greptile.

centrallix-os/sys/js/ht_render.js
Lightning11wins: I think the issue here is that I don't really understand how `eval()` works in `exp_functions.c`. Is there something I should read to understand it better?

Lightning11wins: I think the specific info I'm missing is:
- For `:<property>`, how should I determine what object's property do I default to?
- For `::<property>`, how should I determine who's parent is intended?

Lightning11wins: I think I've resolved this? I'm still not super confident in my understanding, but this implementation looks like it's doing the right thing.

centrallix-os/sys/js/ht_render.js
```diff
@@ -310,6 +310,91 @@ function cxjs_right(s,l)
...
+	    }
+	else if (colon2 < 1)
+	    {
+	    // ::property - parent object reference (C's ParentID)
+	    if (!permflags || permflags.indexOf('P') >= 0)
+		{
+		const prop = rest.substring(1);
+		const par_node = (par_obj_name !== null && par_obj_name !== undefined)
+		    ? wgtrGetNode(_context, par_obj_name)
+		    : wgtrGetParent(_context);
```
gbeeley: This defaults to the parent of the context, rather than the parent of the current object (_this).

In server contexts, the parent object has a context-dependent meaning but usually refers to the immediate parent object of the current object.  The most common use is in a QueryTree object.

On the client, the most useful meaning here would be to reference the parent of _this rather than the parent of _context which likely steps out of the current scope entirely.

centrallix-os/sys/js/ht_render.js
```diff
@@ -310,6 +310,91 @@ function cxjs_right(s,l)
...
+		{
+		if (cur_obj_name !== null && cur_obj_name !== undefined)
+		    {
+		    const cur_node = wgtrGetNode(_context, cur_obj_name);
+		    if (cur_node) return wgtrGetProperty(cur_node, rest);
+		    }
+		else if (_this !== null && _this !== undefined)
+		    {
+		    const v = _this[rest];
+		    return (typeof v === 'undefined') ? null : v;
```
gbeeley: should probably use `wgtrGetProperty(_this, ...)`

gbeeley (changes requested):
A couple of changes needed - thanks!

centrallix-os/sys/js/ht_render.js
Lightning11wins: Ahhh, so like this?

```js
const par_node = (par_obj_name !== null && par_obj_name !== undefined)
    ? wgtrGetNode(_context, par_obj_name)
    : wgtrGetParent(_this);
```

This issue likely arises from me only having a surface-level understanding of what `eval()`, `_context`, and `_this` should do.

centrallix-os/sys/js/ht_render.js
Lightning11wins: Oh, true! Fixed.

centrallix-os/sys/js/ht_render.js
gbeeley: Yes, that looks correct - please test out and verify.  Thanks!

centrallix-os/sys/js/ht_render.js
```diff
@@ -310,6 +310,90 @@ function cxjs_right(s,l)
...
+	    if (!permflags || permflags.indexOf('C') >= 0)
+		{
+		if (cur_obj_name !== null && cur_obj_name !== undefined)
+		    {
+		    const cur_node = wgtrGetNode(_context, cur_obj_name);
+		    if (cur_node) return wgtrGetProperty(cur_node, rest);
+		    }
+		else if (_this !== null && _this !== undefined)
+		    {
+		    return wgtrGetProperty(_this, rest);
```
gbeeley: I take that back.  There are cases where _this is a plain object rather than a widget.  So this code should check first whether it is a widget via `wgtrIsNode()` before calling `wgtrGetProperty()`, otherwise use the _this[prop] approach.

An example of this is parameters passed from an event (the event params) which can be referenced as the current object as in `:property`.

gbeeley (changes requested):
Oops - one of my last recommendations was partly wrong.

centrallix-os/sys/js/ht_render.js
Lightning11wins: I'll try to test this, although I only barely understand what this function is supposed to do in this case.

centrallix-os/sys/js/ht_render.js
Lightning11wins: Hmmm, I think I understand. Should be fixed now.

gbeeley:
@greptileai please recheck - thanks!

centrallix-os/sys/js/ht_render.js
```diff
@@ -310,6 +310,92 @@ function cxjs_right(s,l)
     if (s == null || l == null) return null;
     return s.substr(s.length-l);
     }
+
+function cxjs_eval(_context, _this, expr, permflags, cur_obj_name, par_obj_name)
+    {
+    console.log(_context, _this, expr, permflags, cur_obj_name, par_obj_name);
```
gbeeley: Per greptile, let's remove this debugging line before merge.

gbeeley (changes requested):
Minor fix needed - thanks

centrallix-os/sys/js/ht_render.js
Lightning11wins: Woops! Fixed.

PR #117, opened 2026-06-08
Fix `ORDER BY` Seg Fault (again...)

Lightning11wins:
This PR fixes a segmentation fault caused by some queries that used order by.

Sorry if this gives you de ja vu from #107. It turns out there were more seg faults in that area of the code than we thought. I've done a review, and I'm reasonably sure that I've found all of them.

### DIFF
--- a/centrallix/multiquery/multiq_orderby.c
+++ b/centrallix/multiquery/multiq_orderby.c
@@ -117,7 +117,8 @@ mqobAnalyzeBeforeGroup(pQueryStatement stmt)
 		for(i=0;i<qs->Children.nItems;i++)
 		    {
 		    item = (pQueryStructure)(qs->Children.Items[i]);
-		    if (item->Expr && item->Expr->AggLevel == 0)
+		    if (item == NULL || item->Expr == NULL) continue;
+		    if (item->Expr->AggLevel == 0)
 			{
 			mask = item->Expr->ObjCoverageMask;
 			total_mask |= mask;
@@ -144,7 +145,8 @@ mqobAnalyzeBeforeGroup(pQueryStatement stmt)
 		     ** 2d.  Order item not on the primary side of a join
 		     **/
 		    item = (pQueryStructure)(qs->Children.Items[i]);
-		    if (item->Expr && item->Expr->AggLevel == 0)
+		    if (item == NULL || item->Expr == NULL) continue;
+		    if (item->Expr->AggLevel == 0)
 			{
 			mask = item->Expr->ObjCoverageMask;
 			for(n_sources = 0; mask; mask >>= 1)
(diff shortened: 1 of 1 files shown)


Lightning11wins:
Alright, those changes have been made. This fix isn't trivial anymore.

centrallix/multiquery/multiq_orderby.c
```diff
@@ -414,14 +412,7 @@ mqob_internal_InitSort(pQueryElement qe, pQueryStatement stmt)
...
 	xaInit(&context->AggregateFieldIDs, 16);
-	for(n=0;n<MQ_MAX_ORDERBY;n++)
-	    {
-	    if (!qe->OrderBy[n])
-		{
-		context->nOrderBy = n;
-		break;
-		}
-	    }
+	context->nOrderBy = qe->OrderBy.nItems;
```
gbeeley: We don't need `context->nOrderBy` anymore - just use `nItems` instead.  In the old representation, the count had to be calculated by scanning the array and was then stored for convenience.

gbeeley (changes requested):
One minor change needed to clean up.  Thanks!

centrallix/multiquery/multiq_orderby.c
Lightning11wins: Good to know, the dead code has been removed.

Lightning11wins:
**TODO**: Revert changes to object system contract.

Lightning11wins:
**Done**, I have reverted changes to the object system contract and affected drivers.

Lightning11wins:
This reintroduced the 16 item order by and group by limits: this is now noted in issue #165.

gbeeley:
Huh, I had only intended for the OSML->OSD interface to be reverted, meaning the drivers still take a plain array, but not revert the change to QueryElement's order by and group by arrays.  Those can still be XArrays.  The main issue was to defer changes which would require a modification of the database OSDs.

Lightning11wins:
According to Claude, the Sybase driver implements both the OSML->OSD interface (with `ObjQuery.SortBy`) *and* the multiquery query-driver interface (`QueryElement.OrderBy`).

I find the idea that the sybase driver implements the multiquery query-driver interface unintuitive and confusing. Perhaps it's because I don't fully understand the interfaces here. From reading `centrallix/osdrivers/objdrv_sybase.c:4889`, it *looks* like it's relying on `QueryElement.OrderBy`, hence my decision to revert changes to that data structure.

gbeeley:
That QE interface in sybd wasn't in use - it was an early experiment for passing joins through to the DB.  That concept was abandoned for various reasons some of which we've discussed.  But nevertheless, you are right that modifying the QE structure would force an edit to sybd, so yes I think the direction you went with this is sound.

gbeeley (approved):
Looks good - thanks

PR #118, opened 2026-06-10
Fix Textarea Font Size

Lightning11wins:
This quick PR fixes #115, a bug that caused the font size of textarea widgets to be too large.

### Before
Example Image

### After
image


gbeeley (approved):
Looks good - thanks!

PR #122, opened 2026-06-10
Update `cxjs_substring()` to match `exp_fn_substring()`

### LINKED ISSUE #121: Inconsistent `cxjs_substring()` Implementation
`substring("1234", 0, 1)` produces different results on the client (`"4"`) than on the server (`"1"`).

Lightning11wins:
I caused a bug in the `gift_associations.app` because the client-side version of `substring()` functioned differently from the server side version, as documented in #121. This PR will close #121 by solving this inconsistency issue.

### DIFF
--- a/centrallix-os/sys/js/ht_render.js
+++ b/centrallix-os/sys/js/ht_render.js
@@ -300,10 +300,19 @@ function cxjs_convert(dt,v)
 function cxjs_substring(s,p,l)
     {
     if (s == null || p == null) return null;
-    if (l == null)
-	return (String(s)).substr(p-1);
-    else
-	return (String (s)).substr(p-1,l);
+    if (typeof s != 'string' || !Number.isInteger(p) || (l != null && !Number.isInteger(l)))
+	return null;
+    
+    let i = p - 1;
+    if (i < 0) i = 0;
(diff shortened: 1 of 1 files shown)


centrallix-os/sys/js/ht_render.js
```diff
@@ -300,10 +300,19 @@ function cxjs_convert(dt,v)
 function cxjs_substring(s,p,l)
     {
     if (s == null || p == null) return null;
-    if (l == null)
-	return (String(s)).substr(p-1);
-    else
-	return (String (s)).substr(p-1,l);
+    if (typeof s != 'string' || !Number.isInteger(p) || (l != null && !Number.isInteger(l)))
```
gbeeley: Note that for String objects, typeof returns 'object', not 'string'.  Both string objects and string primitives need to work in these functions (for compatibility reasons).

gbeeley (changes requested):
Change needed for string object compatibility.

centrallix-os/sys/js/ht_render.js
Lightning11wins: Oh, I see. Fixed.

gbeeley (approved):
Looks good - thanks

PR #123, opened 2026-06-10
Fix Namespace Bug

### LINKED ISSUE #120: `hints` Widgets Applied to `component` Widgets Use Incorrect Namespace
Consider the following simplified scenario, in a `.app` file:
```centrallix
gift_uuid "widget/variable" { fieldname = i_eg_gift_uuid; }

gift_id_field "widget/component"
    {
    x = 0; y = 0; width = 180; height = 15; label_width = 60;
    path = "/sys/cmp/smart_field.cmp";
    text = "Gift ID:";
    ctl_type = label;
    
    gift_id_hints "widget/hints"
	{
	default = runclient(condition(
	    char_length(:gift_uuid:value) > 12,
	    substring(:gift_uuid:value, 0, 12) + "...",
	    :gift_uuid:value
        ));
	}
    }
```

In this scenario, the expression in `gift_id_hints` is not able to access `gift_uuid` because it runs in the scope of `smart_field.cmp`, not in the scope of the `.app` file, breaking the app.

Lightning11wins:
This PR fixes #120, so hints widgets applied in components now use the correct namespace context. Thus, this PR closes #120.

### DIFF
--- a/centrallix-os/sys/js/ht_utils_hints.js
+++ b/centrallix-os/sys/js/ht_utils_hints.js
@@ -1,4 +1,4 @@
-// Copyright (C) 1998-2004 LightSys Technology Services, Inc.
+// Copyright (C) 1998-2026 LightSys Technology Services, Inc.
 //
 // You may use these files and this library under the terms of the
 // GNU Lesser General Public License, Version 2.1, contained in the
@@ -37,8 +37,9 @@ cx_hints_style.key = 262144;
 cx_hints_style.applyonchange = 524288;
 
 // cx_set_hints() - initializes hints information for a given
-// form field.
-function cx_set_hints(element, hstr, hinttype, applydef)
+// form field.  'context' preserves the namespace where hints
+// expressions were authored when hints are applied in components.
+function cx_set_hints(element, hstr, hinttype, applydef, context)
     {
     if (!element.cx_hints) element.cx_hints = {};
     if (element.cx_hints.hstr == hstr) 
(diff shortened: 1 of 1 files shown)


centrallix-os/sys/js/ht_utils_hints.js
```diff
@@ -158,6 +161,7 @@ function cx_merge_two_hints(h1,h2)
 	if (!h2) return h1;
 	nh.Constraint = cx_merge_hint_expr(h1.Constraint, h2.Constraint, 'cx_AND');
 	nh.DefaultExpr = cx_merge_hint_expr(h1.DefaultExpr, h2.DefaultExpr, 'cx_FIRST');
+	nh.Context = h1.DefaultExpr ? h1.Context : (h2.DefaultExpr ? h2.Context : null);
```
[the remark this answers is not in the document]
Lightning11wins: Fixed.

centrallix-os/sys/js/ht_utils_hints.js
```diff
@@ -175,6 +178,20 @@ function cx_merge_two_hints(h1,h2)
...
 	nh.GroupName = cx_merge_hint_string(h1.GroupNAme, h2.GroupName);
 	nh.FriendlyName = cx_merge_hint_string(h1.FriendlyName, h2.FriendlyName);
+	
+	// Note: If both layers carry a DefaultExpr from different namespaces,
+	// we can't safely combine them with cx_FIRST.  Thus, we keep h1
+	// (which already wins under cx_FIRST priority) and drop h2's default.
+	if (h1.DefaultExpr && h2.DefaultExpr && h1.Context !== h2.Context)
+	    {
+	    nh.DefaultExpr = h1.DefaultExpr;
+	    nh.Context = h1.Context;
```
gbeeley: This logic doesn't account for the situation where both hint sources have a valid DefaultExpr, but h1's expression returns null while h2's expression does not.  In that case the value would be set to null and h2's expression ignored completely.  We need some mechanism here to merge two hint expressions with different scopes and keep the scopes/contexts intact for both.

Incidentally this situation applies also to other expression based hints.  We probably need a construct like `cx_eval_in_scope(scope, expression)` that can be placed within expression strings.

gbeeley (changes requested):
logic change needed for default expressions - thanks!

centrallix-os/sys/js/ht_utils_hints.js
Lightning11wins: I just barely understand how this code works at all, so I'm not sure how to design that in a way that works well with the rest of this system. We might need to have a design discussion.

gbeeley: I think the cx_eval_in_scope(scope, expression) approach is the best approach here.  It would require that new function that explicitly specifies the scope an expression should be evaluated in.  Then, the DefaultExpr handling should evaluate h1, and if it is unspecified or the result is null, evaluate h2 similarly.  Scopes do have global identifiers and so can be specified directly.

Lightning11wins: Ok, I'll give it a shot and we'll see what comes out.

Lightning11wins: Just pushed my and Claude's best attempt to implement this despite my minimal understanding of this part of the system.

Lightning11wins:
@gbeeley Please let me know what you think of this implementation.

gbeeley:
@greptileai please re-check with the updated hint merging logic.

gbeeley (approved):
Looks good - thanks

PR #124, opened 2026-06-10
Fix Sort Being Lost on `osrc` Refresh

Lightning11wins:
Refreshing a sorted osrc causes the sort to be lost. This happened for me when I tested refreshing content in a table after the user had clicked a column to sort the content, and the Refresh action dropped the sort. This PR fixes that.


gbeeley (approved):
Looks good - thanks!

PR #125, opened 2026-06-11
Fix Appending `ORDER BY` Clauses

Lightning11wins:
This PR will fix osrc widgets appending `ORDER BY` clauses to queries that already have an `ORDER BY` clause.

This is most commonly seen a table uses an osrc with a query that contains an `ORDER BY` clause. When the user clicks a column in the table header row to sort by that column, the user's new order is appended to the query and subsequently ignored by the parser, which only parses the first clause.


gbeeley (changes requested):
See note in conversation.  I think there is a much better way to solve this problem.

PR #126, opened 2026-06-11
Add `allow_sorting` Property

Lightning11wins:
This PR adds the `allow_sorting` property to `table-column` widgets that allow the app designer to disable clicking the header row of a table to sort by that given column. It also adds the `allow_sorting` property to the entire table, where it serves as a default for columns where the property is not specified. This value defaults to `yes`, which gives the current behavior.

This functionality will be very useful for apps where clicking the header row to sort is either not needed, does not make sense, or is just simply very hard to implement correctly.


gbeeley (approved):
Looks good - thanks!

PR #127, opened 2026-06-17
Handle Multiple `ORDER BY`s with `DEFAULT`s

Lightning11wins:
This PR will add functionality for handling multiple `ORDER BY` statements. Later statements append to earlier ones, or overwrite them if they use the `DEFAULT` keyword, as [suggested by Greg](https://github.com/LightSys/centrallix/pull/125#issuecomment-4723094560).

This PR includes both docs and testing for the new functionality and the `DEFAULT` keyword.


centrallix/multiquery/multiquery.c
```diff
@@ -640,6 +642,59 @@ mq_internal_PostProcess(pQueryStatement stmt, pQueryStructure qs, pQueryStructur
...
+	     /** Skip non ORDER BY clauses. **/
+	    if (subtree->NodeType != MQ_T_ORDERBYCLAUSE)
+		{
+		i++;
+		continue;
+		}
+
+	    /** Drop a DEFAULT clause that isn't the last one. **/
+	    if ((subtree->Flags & MQ_SF_DEFAULTORDER) && !is_last)
+		goto remove_order_by_element;
```
gbeeley: Avoid the goto since this isn't a retry mechanism or an error condition handler, and since this can be easily refactored as an if() block.

centrallix/multiquery/multiquery.c
```diff
@@ -1441,6 +1496,18 @@ mq_internal_SyntaxParse(pLxSession lxs, pQueryStatement stmt, int allow_empty, p
...
+				/*** Optional DEFAULT keyword: Marks this ORDER BY
+				 *** to be replaced by later ORDER BYs, instead of
+				 *** appended to. (see mq_internal_PostProcess()).
+				 ***/
+				if ((t = mlxNextToken(lxs)) == MLX_TOK_KEYWORD &&
+				    (ptr = mlxStringVal(lxs, NULL)) != NULL && 
+				    strcasecmp(ptr, "default") == 0)
+				    orderby_cls->Flags |= MQ_SF_DEFAULTORDER;
+				else
+				    mlxHoldToken(lxs);
```
gbeeley, line 1514: Add "default" to the list of reserved words at the top of this function, check type against MLX_TOK_RESERVEDWD, and do the string compare using strcmp() (in this parse, mtlexer is set to automatically lowercase reserved words).

gbeeley (changes requested):
A couple of changes needed - thanks!

centrallix/multiquery/multiquery.c
Lightning11wins: `goto` removed.

centrallix/multiquery/multiquery.c
Lightning11wins: I see, updated.

Lightning11wins:
This PR no longer seems to be of trivial size.

Lightning11wins:
> Let's resolve PR 117 first, moving to XArray, and then revisit this one.

The move to XArray for OrderBy in PR #117 has been completed.

gbeeley (approved):
Looks good - thanks!

PR #132, opened 2026-07-24
Reduce Optimization Warnings

Lightning11wins:
Reduce more warnings that only occur when `--enable-optimization` is used. Some of these warning fixes also fix minor bugs and improve hardening.

This PR is blocked behind #95.

### DIFF
--- a/centrallix-lib/src/mtask.c
+++ b/centrallix-lib/src/mtask.c
@@ -694,7 +694,10 @@ int
 r_mtRun_PokeStack()
     {
     char buf[MT_MAX_STACK - MT_TASKSEP*2];
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wuninitialized"
     buf[0] = buf[0];
+#pragma GCC diagnostic pop
     return 0;
     }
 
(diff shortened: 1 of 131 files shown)


centrallix/objectsystem/obj_datatypes.c
```diff
@@ -1958,24 +1958,26 @@ objDataToWords(int data_type, void* data_ptr)
 		    m->WholePart, m->FractionPart
 		);
 		}
+
+	    /** Clamp the cents to the low four digits **/
+	    fraction_part = m->FractionPart % 10000;
```
gbeeley, line 1963: This will be better resolved once we merge the MoneyType representation PR, but I think this change makes the problem worse, though the warning message helps.  It prevents weird output from a malformed MoneyType value, but it does so by hiding the malformed data and making it look like everything's okay.  It'd be better to either have it print a malformed value weirdly, fold the out of range `fraction_part` into the `integer_part`, or else to return an error.  The latter may be best here.

gbeeley (changes requested):
Fix needed for modulus clamping of `fraction_part`.

centrallix/objectsystem/obj_datatypes.c
Lightning11wins: Alright, switching to returning an error.

Lightning11wins: Done.

gbeeley (approved):
Looks good - thanks

PR #133, opened 2026-07-28
Add Style & Linting

Lightning11wins:
Add a new, more detailed style guide documenting the new Beeley style in far more detail than the previous document.

**TODO**: Configure and run a linter with these new styles.


centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+
+
+## Style Rules
+All code should follow a consistent style when possible.  This helps developers to read it quickly and accurately without being tripped up by abrupt changes in style or unfamiliar coding practices.
+
+These rules apply to every language in Centrallix, unless a section states otherwise.  A section that only covers certain file types is marked with a `For:` line listing them.  Examples use C, the most common language in the codebase.
+
+
+### Indentation
+- Code is always indented using 1 tab, not with spaces.
```
nboard: **Indentation**
Consider, `Each level of indentation should use a single tab. Do not use spaces for general indentation.`

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- You can set your editor tab length to any size you prefer.  However, code should look reasonable with any tab length up to 8 spaces.
+- Prefer tabs, but always use spaces for aligning characters, such as in a copyright notice or a column of struct members.  Tabs break alignment when the viewer uses a different tab length.
+- Lines should not have trailing whitespace.  Thus, blank lines are not indented.
+- In a makefile, the tab that begins a recipe line is required by the syntax, so it must be a literal tab.  Replacing it with spaces breaks the build.
```
nboard: **Indentation**
Consider making this statement more general, like, `Language syntax that requires literal tabs and/or spaces must be followed`, to avoid getting into the weeds on the syntax of other scripting languages. Or did you think the specific case of makefile recipes was likely to trip people up?

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- Code is always indented using 1 tab, not with spaces.
+- You can set your editor tab length to any size you prefer.  However, code should look reasonable with any tab length up to 8 spaces.
+- Prefer tabs, but always use spaces for aligning characters, such as in a copyright notice or a column of struct members.  Tabs break alignment when the viewer uses a different tab length.
+- Lines should not have trailing whitespace.  Thus, blank lines are not indented.
+- In a makefile, the tab that begins a recipe line is required by the syntax, so it must be a literal tab.  Replacing it with spaces breaks the build.
+
+### Spacing
+- Language constructs (e.g. `if ()`, `for ()`, `while ()`, `switch ()`, etc.) should always be separated from their parentheses with a space.
+- Function calls (e.g. `mssError()`) and function declarations should *not* be separated from their parentheses by a space.
+- Pointer types attach the `*` to the type, not to the identifier: `char* p`, not `char *p`.  A pointer is part of the type, so it belongs with the type.
```
nboard, line 53: **Spacing**
If we do this, we also need to only allow one pointer declaration per line, as `char* a, b;` and `char* a, *b;` are both undesirable for different reasons.

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+
+### Indentation
+- Code is always indented using 1 tab, not with spaces.
+- You can set your editor tab length to any size you prefer.  However, code should look reasonable with any tab length up to 8 spaces.
+- Prefer tabs, but always use spaces for aligning characters, such as in a copyright notice or a column of struct members.  Tabs break alignment when the viewer uses a different tab length.
+- Lines should not have trailing whitespace.  Thus, blank lines are not indented.
+- In a makefile, the tab that begins a recipe line is required by the syntax, so it must be a literal tab.  Replacing it with spaces breaks the build.
+
+### Spacing
+- Language constructs (e.g. `if ()`, `for ()`, `while ()`, `switch ()`, etc.) should always be separated from their parentheses with a space.
```
nboard, line 50: **Spacing**
Could we simplify things by not having a space (which makes it match function calls)? It looks like the current code is pretty split on this, so I think any decision (as long as function calls don't have a space) wouldn't be unreasonable.

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+
+### Spacing
+- Language constructs (e.g. `if ()`, `for ()`, `while ()`, `switch ()`, etc.) should always be separated from their parentheses with a space.
+- Function calls (e.g. `mssError()`) and function declarations should *not* be separated from their parentheses by a space.
+- Pointer types attach the `*` to the type, not to the identifier: `char* p`, not `char *p`.  A pointer is part of the type, so it belongs with the type.
+- Do not put a space just inside parentheses or brackets.
+- Commas and semicolons are followed by a space (or line break), never preceded by one.
+- Unary operators, `->`, `.`, `[]`, and casts are written against their operand, e.g. `!found`, `&data`, `node->Name`, `list[i]`, `(char*)ptr`.
+- In a makefile, an assignment is spaced like any other binary operator, e.g. `CC = @CC@`.
```
nboard: **Spacing**
Consider, `Binary operators should have a space on either side of the operator. This includes assignment in make files, e.g. ...`
You could add the note about make files as a sub-bullet if desired.

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- Function calls (e.g. `mssError()`) and function declarations should *not* be separated from their parentheses by a space.
+- Pointer types attach the `*` to the type, not to the identifier: `char* p`, not `char *p`.  A pointer is part of the type, so it belongs with the type.
+- Do not put a space just inside parentheses or brackets.
+- Commas and semicolons are followed by a space (or line break), never preceded by one.
+- Unary operators, `->`, `.`, `[]`, and casts are written against their operand, e.g. `!found`, `&data`, `node->Name`, `list[i]`, `(char*)ptr`.
+- In a makefile, an assignment is spaced like any other binary operator, e.g. `CC = @CC@`.
+- The `return` statement should be followed by a space and the return value should only use parentheses when needed.  Do not treat `return` as a function call, it is not a function, and it does not give a return value... well, it kind of does, but I think you get the point :)
```
nboard, line 59: **Spacing**
Consider trimming this to just the first sentence.

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+
+### Braces
+- Braces use a style similar to the Linux kernel.
+	- Braces are placed on the same line as the initiating structure.
+	- Braces are terminated at the same indentation as the initiating structure.
+- Any `if` statement, `for` loop, `switch` statement, or similar structure must use braces if it contains:
+	- More than one line of code.
+	- Another such structure.
+- An `else` or `else if` starts on the line of the closing brace of the block before it, not on the line after that brace.
```
nboard, line 69: **Braces**
Consider adding an example here. If a reader was confused by either the first line or this one, an example here could clear it up.

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- **Exception**:  In a structure file, braces group data rather than code, so every group is braced regardless of its number of attributes.
```
nboard, line 81: **Braces**
I believe this is required by the language syntax, so you may not need to state it explicitly here.

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+		id "system/querypivot" { type=integer; usage=key; }
+		```
+
+### Switch Statements
+- `case` and `default` labels are indented one level inside the `switch`.
+- Every case body is enclosed in braces, even a body of one line, and the `break` goes inside those braces.
+- Fallthrough is allowed, but it must be explicitly marked with a `/** Fallthrough. **/` comment.
+
+For example:
+```c
```
nboard, line 93: **Switch Statements**
Consider using a simpler toy example to avoid bogging down the reader with long constant names and function calls.

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- Single line comments not to the right of code (including section comments) use double asterisks, e.g. `/** this is a single line comment **/`.
+- Multiline comments (including function comments) use three asterisks, which continue on each line, e.g.
+	```c
+	/*** myFunFunction - this is a function
+	 *** that does some fun stuff with its
+	 *** parameters.
+	 ***/
+	```
+- In a `.c` or `.h` file, never use a `//` comment.  C99 permits it, but it does not match with the asterisk tiers above.
+- **Exception**:  A file may only use the comment syntax its language accepts, such as `//` in an `.app` file, `#` in a makefile, or `<!-- -->` in XML.  Follow the rules above to the extent that syntax allows.  XML has no single-line comment syntax and cannot nest comments, so the asterisk tiers above do not apply to it at all.
```
nboard: **Comments**
This is a good point. What do we want JS comments to look like? right `// coment`, single `/// comment`, multi line 
```javascript
//// myFunction
//// params and such
//// aaand done
```
Or should we have them use `*` symbols instead? Drop the hierarchy altogether outside of C?
Query files, app files, and make files can probably all use the same solution we land on for JS

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+	```
+- In a `.c` or `.h` file, never use a `//` comment.  C99 permits it, but it does not match with the asterisk tiers above.
+- **Exception**:  A file may only use the comment syntax its language accepts, such as `//` in an `.app` file, `#` in a makefile, or `<!-- -->` in XML.  Follow the rules above to the extent that syntax allows.  XML has no single-line comment syntax and cannot nest comments, so the asterisk tiers above do not apply to it at all.
+
+### Include Files
+For: `.c`, `.h`
+
+- If a `.c` file uses an identifier from a `.h` or C standard library not available by default, that file must `#include` the correct `.h` file or library.
+- For example, the `mssError()` function is defined in `mtsession.h`, so that file must have a `#include` in every C file that uses `mssError()`
+	```c
```
nboard: **Include FIles** 
If you are looking to trim things down, I think this example is less helpful than some of the others

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- A directive at the outermost level starts in the first column, even inside a function.
+- Nested directives are indented one space per level of nesting, placing the spaces before the `#`.  A `.h` file's [include guard](#include-files) does not count as a level, since it wraps the whole file.
+	```c
+	#ifdef HAVE_CONFIG_H
+	 #ifndef MOD_MAX_ITEMS
+	  #define MOD_MAX_ITEMS 256
+	 #endif
+	#endif
+	```
+- A `#define` that spans multiple lines ends every line but the last with a backslash, placed one space after that line's content.  Do not align the backslashes in a column because one long line added later would force a reflow of every line.
```
nboard: **Preprocessor Directives**
Consider splitting the comment about not aligning the `\` into a separate bullet or note

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- Continuation lines are indented one level.  A comment may take a continuation line of its own.
+	```c
+	#define PRT_HTMLFM_EMAIL_CONTENT_HEADER "\n" \
```
nboard: **Preprocessor Directives**
Consider a simpler/toy example

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+		"\n"
+	```
+
+### Naming Identifiers
+All identifiers should be spelled correctly and avoid using non-obvious abbreviations.  The very common `i`, `j`, and `rval` local variables are exceptions to this general rule.
+- Functions are named with camelCase.
+	- The module prefix is prepended if the function is reachable from outside the module. (The prefix is optional for internal & unreachable functions.)
+	- Internal functions prepend `_internal_` or `_i_`. (e.g. `xyz_i_hiddenStuff()`)
+- Local variables and function parameters are named with snake_case.
+- Every global variable in a module lives in a single module-wide global struct named with SCREAMING_SNAKE_CASE with the module name prepended (for easy identification).  The recommended name is `MOD_GLOBALS` (where `MOD` is the module prefix).
```
nboard: **Naming Identifiers**
Consider `XXX_GLOBALS` as a dummy value that stands out a bit more as fake

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- `Typedef` names & structs:
+	- `Typedef`ed names use PascalCase (except for the module prefix).
+	- Both the struct type (`XxxxYyy`) and a pointer alias (`pXxxxYyy`) are declared in the same typedef.
+	- Structs reachable from outside the module start their name with the module prefix.  (As with functions, the prefix is optional for internal structs.)
+	- Members of a struct or union use PascalCase.
+		- **Exception**:  A count member may take a lowercase `n` prefix, and a pointer member a lowercase `p` prefix, before the PascalCase name.  (e.g. `nData`, `pCluster`)
+- In C, value macros are treated as globals and follow the naming style of the global struct.
```
nboard: **Naming Identifiers**
Consider, `In C, value macros are treated as globals, and use SCREAMING_SNAKE_CASE with a module prefix for any values used outside of the module.`

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+
+## Style Rules
+All code should follow a consistent style when possible.  This helps developers to read it quickly and accurately without being tripped up by abrupt changes in style or unfamiliar coding practices.
+
+These rules apply to every language in Centrallix, unless a section states otherwise.  A section that only covers certain file types is marked with a `For:` line listing them.  Examples use C, the most common language in the codebase.
+
+
+### Indentation
+- Code is always indented using 1 tab, not with spaces.
+- You can set your editor tab length to any size you prefer.  However, code should look reasonable with any tab length up to 8 spaces.
```
nboard: **Indentation**
I think we could gain some advantages from having a recommended width, which would solve the issue with not being able to use tabs to align rows in structs and the like. Granted, we would probably end up with a tab space of 4, which would not not simplify such alignments by much...

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+If a struct supports `magic.h` by beginning with a magic field of type `Magic_t`, the following rules should be followed:
+- When creating the struct, call `SETMAGIC()` with the appropriate magic value (e.g. `MGK_FILE`).
+- When a new scope first gains access to the struct (e.g. the first time a function reads and stores a pointer to it) and intends to read any field from it (rather than just passing the pointer to another scope), it must call `ASSERTMAGIC()` immediately (after verifying that the struct is not null, if needed).  No data should be read from the struct or used before `ASSERTMAGIC()` is called.
+
+### File Organization
+For: `.c`, `.h`, structure files
+
+In a structure file, `$Version=2$` is always the first line, placed before the copyright notice (optional for files in `centrallix-os`).
+
+In a `.c` or `.h` file, it is recommended to order the top level of the file as follows, for consistency with other files, unless there is a good reason to do otherwise (such as a forward reference):
```
nboard: **File Organization**
Consider simplifying the introduction to just,  `... file, it is recommended to order the top level of the file as follows:`, and then add the exception for forward references as an additional note.

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+1. `#include` guards (in a `.h` file).
+2. Copyright notice.
+3. `#include` groups.
+4. Macros.
+5. `Typedef`s, structs, and unions.
+6. Prototypes for functions defined later in the same file.
```
nboard: If it is a .h file, there would be prototypes but no definitions, right?

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+For: `.c`, `.h`, structure files
+
+In a structure file, `$Version=2$` is always the first line, placed before the copyright notice (optional for files in `centrallix-os`).
+
+In a `.c` or `.h` file, it is recommended to order the top level of the file as follows, for consistency with other files, unless there is a good reason to do otherwise (such as a forward reference):
+1. `#include` guards (in a `.h` file).
+2. Copyright notice.
+3. `#include` groups.
+4. Macros.
+5. `Typedef`s, structs, and unions.
```
nboard: The formating on the s looks odd... Consider "Typedefs", "`Typedef` statements", or maybe change the whole line to, "`Typedef`, `struct` and `union` declarations"

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+6. Prototypes for functions defined later in the same file.
+7. Function definitions (`.c` files).
+
+**Note**: Many files may not have all of these sections.
+
+### Types
+For: `.c`, `.h`
+
+- C code should work for any C of C99 or later, so `<stdbool.h>` and variable declarations inside a `for` statement are always available.
```
nboard: **Types**
Consider, "C code should work for version C99 or later..."

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+	const unsigned int max_iter,
+	const double min_improvement,
+	unsigned int* labels,
+	double* vector_sims,
+	bool auto_seed
+) {
+	...
+}
+```
+**Note**: This should be a rare occurrence because, in many cases, it's better to avoid functions with large numbers of parameters.  Calling such functions requires extra care to avoid confusing calls like: `caKMeans(v, 32, 256, 12, 0.01, &l, &s, false);`.  If large numbers of parameters are truly the best solution in your situation, using clear variable names for even *some* parameters, as opposed to passing magic numbers, reduces this issue, e.g.: `caKMeans(vectors, n_vectors, n_clusters, n_iters, 0.01, &labels, &sims, false);`
```
nboard, line 358: **Function Declarations**
Is it worth explicitly noting that creating a struct to hold the parameters and passing that is a valid option, or is that obvious, you think?

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+
+### Function Calls
+*When in doubt, use similar styling to [function declarations](#function-declarations).*
+
+- Short function calls should also be placed on one line: `getTime(&timer)`.
+- `sizeof` is an operator, not a function, but it is styled as a function call: write `sizeof(ModDataT)`, with no space before the parenthesis.
+- Function calls with many or long parameters may span multiple lines.  Literal parameters (especially string literals) may even be split across lines.  Situations vary, so use discernment and consistency to write the most clear code.  The most common example of this is the C calls to write inline `HTML` and `CSS`:
+	```c
+	htrAddStylesheetItem_va(s,
```
nboard: **Function Calls**
In a function declaration, the first parameter is on the line below the function name, but when called here, the first argument is on the same line as the name. I would probably be inclined to make those same decisions without thinking, but I wonder if we should standardize that here by making the declarations and calls match?

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+			"visibility:hidden; "
+			"left:0px; "
+			"top:0px; "
+			"width:"ht_flex_format"; "
+			"height:%POSpx; "
+			"z-index:%POS; "
+		"}\n",
+		t->id, detail_id,
+		ht_flex_w(t->w, tree),
+		h,
```
nboard: **Function Calls**
I find it a bit odd that `t->id` and  `detail_id` share a line, but `h` and `z + 1` each get their own lines. I think this particular example would benefit from a longer explanation behind the seaming inconsistencies.

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+		z + 1
+	);
+	```
+	- Unsurprisingly, writing such calls clearly can be a real struggle!
+
+### Local Variables
+- Local variables should be declared as close to where they are used as possible, and especially within the narrowest scope.  This reduces the amount of worrying about side effects and scrolling around that a programmer must do when seeing them.
```
nboard: **Local Variables**
Consider, "... This reduces the risk of potential side effects and allows the programmer to scroll around less when reading the code."

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- Variables should be declared using `const` (or the equivalent keyword for the relevant language) when their value is not changed.  This serves as a note to the reader so they know that the value won't change.
+- In `js`, declare with `const`, or with `let` when the value changes.
```
nboard: **Local Variables**
Consider, "in `js`, declare variables with `let` when the value changes. Otherwise, use `const`."

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+		"}\n",
+		t->id, detail_id,
+		ht_flex_w(t->w, tree),
+		h,
+		z + 1
+	);
+	```
+	- Unsurprisingly, writing such calls clearly can be a real struggle!
+
+### Local Variables
```
nboard, line 396: **Local Variables**
I think we should also recommend that they not reuse the same variable for multiple purposes. Many functions have a `char* ptr;` declared at the start which is used to store various function results with very different purposes. This would be more resonable if there were multiple variables with the same name limited to local scopes, since the potential for side effects is lessened. Ideally they would just use unique variables with descriptive names to avoid confusion 

We should probably also mention that any variables that contain data that will need freed on error or at the end of the function need to be allocated at the start of the function and have function wide scope (so we don't end up with code trying to determine if a variable that was potentially never even declared needs freed).

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- Incorrect example:
+	```c
+	if (source->Name) printName(source->Name);
+	if (!strcmp(source->Title, "None")) queryTitle(source->Title);
+	```
+
+### Equality
+For: `.js`
+
+- Compare with `===` and `!==`, not `==` and `!=`.  The loose operators convert their operands before comparing, which surprises readers and hides bugs.
```
nboard: **Equality**
I feel like it might be easy to miss the "not" when skimming the list. Consider, "Compare with `===` and `!==`. Do not use `==` and `!=` as the loose operators..."

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- **Exception**:  `x == null` (or `x != null`) as an idiomatic way to test for `null` and `undefined` together is allowed.
+
+### XML
```
nboard: **XML**
I feel like this should be later in the file rather than separating topics that pertain to C code more

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+### File Encoding
+- All files should be encoded using UTF8.
+- All files should have LF line endings.
+- All files should end with a newline, so the last line of content is complete (this also makes writing to the end easier).
+
+### Constant Sets
+For: `.c`, `.h`
+
+A constant set is a named group of numerical values, such as the algorithms a module supports or the flags a structure carries.  Every set follows these rules:
+- The set should have a `typedef` of an unsigned numerical type, so declarations using this type will show what the value means.
```
nboard: **Constant Sets**
Ideally, the the declaration would have a comment explaining what the set was for, so I feel like this does not benefit the declaration in that way. I do think typecasting is useful for removing ambiguity with how the compiler will treat the constant.

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+	- The underlying numerical type should be the minimum size necessary for the number of values defined.
+	- Valid numerical types include at least: `unsigned char`, `unsigned short`, `unsigned int`, `unsigned long`, and possibly others.  Unsigned types are required to prevent signed issues.
```
nboard: **Constant Sets**
There are actually examples of constant sets that intentionally use negative values. They could be changed to unsigned types and just use the corresponding large values, but I thought it worth mentioning. 
This regex should work to find examples in the code base, if you're curious: `#define.*-[0-9]`

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- Each value is defined as a macro, cast to the set's type with an unsigned literal.  The cast carries the type into comparisons and `switch` statements, which a bare number would not.
+- Values are aligned in a column (see [struct & union declarations](#struct--union-declarations) for the alignment rules, which apply here too).
+- Macro names use the scheme `MOD_SET_XXX`, where "`MOD`" is the module prefix, "`SET`" is a short abbreviation for the set, and "`XXX`" is the name of the individual value.
```
nboard: **Constant Sets**
While often tied to a struct, I would argue the flag name should more so reflect the module it is a part of and the purpose it serves within the module.

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+	#define CA_ALG_DB_SCAN          ((ClusterAlgorithm)6u)
+	```
+
+#### Flags
+- A flag set can hold any number of its values at once, so each value is a distinct power of two.
+- Flags are always stored in full-length integers, and never as single-bit bitfields (i.e., `Flag:1;`).  This allows bulk editing and passing multiple flags as one parameter.
+- Flag names insert an `F` tag: `MOD_STRUCT_F_XXX`, where "`STRUCT`" is the structure the flags serve.
+- Flag comparisons should use `flags & MOD_STRUCT_F_XXX`, which is considered a boolean for the purposes of the [Truthiness](#truthiness) rule.
+
+#### Multi-Type Structures
```
nboard: **Constant Sets**
The first two bullets seem like they pertain more to the struct definition than the constants. Consider either moving the requirements to the section on struct requirements, or modifying them to be more descriptive than imperative.

centrallix-sysdoc/CentrallixCodingStyle.md
nboard, line 1: I believe the name of the file could be more descriptive, and I feel this document should replace the old Beeley coding style document. Consider creating a single "centrallix coding style" document.

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- Commented-out code should be deleted before code is reviewed (preferably before it is even committed).
+- Unreachable code should be removed, or clearly marked in the rare case that immediate removal is not possible.
+	- This does not apply to code that only *happens* to be unreachable.  For example, the default case on a switch statement with cases for all defined values isn't considered "unreachable" here if the list of values is owned by another module.  In fact, this default case (usually an error case) would serve as a valuable warning if a new possible value was defined and the switch statement is not updated.
+
+### File Encoding
+- All files should be encoded using UTF8.
+- All files should have LF line endings.
+- All files should end with a newline, so the last line of content is complete (this also makes writing to the end easier).
+
+### Constant Sets
```
nboard: **Constant Sets**
With how closely related to the **Naming Identifiers** and **Struct & Union Declarations** sections this is, I feel like it should be closer to them

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+			/** Some logic... **/
+		} else goto error;
+	} else goto error;
+} else goto error;
+```
+
+Typically, errors in C code will `goto` an error handler at the end of the current scope or function.  This code is responsible for clearing up memory.  It also usually logs an error message with any helpful info available in that scope so that individual error sites don't need to specify that info.
+
+In C, the error handler follows these rules:
+- The handler is labeled `error:` or `end:` (if it also handles a success case), placed at the end of the function or scope, and indented only once (typically the level of the function body).
```
nboard: **Full Error Handling Example**
In the current code style the tags are indeed indented once, but the body is indented twice so the tags still stand out. I think we should have the tags match the level of the function's `{}`, which would mean leaving it un-indented

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
+# Beeley's Coding Style 2
+
+**Author**: Israel Fuller
+
+**Date**: June 24th, 2026
+
+**License**: Copyright (C) 2026 LightSys Technology Services.  See `LICENSE`.
+
+
+## Introduction
```
nboard, line 12: I think this document would benefit from a table of contents eventually (though, it's probably best to wait until the document is more stable before adding one, as the upkeep will be a pain otherwise)

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+When you gracefully handle an error, if `mssError()` or similar functions are called (or likely to be called) during an error, call `mssClearError()` to mark the error as handled.  This prevents messages from this error from appearing at the start of the error stack for a later error.
+
+
+## Comments & Documentation
+Code should use a few types of documentation, where applicable.  See [comments](#comments) for commenting style.
+
+Any English sentence should use two spaces (or a newline/similar whitespace) between a period and the following sentence.  This makes it easy to visually jump between sentences and skim text.  This is not required for string literals, such as error messages in code, although these rarely have more than a single sentence anyway.
+
+### Warning Comments
+- Unintuitive warnings and "foot-guns" should be called out in brief, concise comments.
```
nboard: **Warning Comments**
Consider, "Warnings for unintuitive code and..."

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- These should be very rare, only used if there isn't a better way to communicate the information.
+
+### Section Comments
+- Divide code into logical sections that start with brief comments for skimming code.
+- Sections should be separated by blank lines.
+- These comments should start with a capital letter and end with a period.
+- These comments should be concise, only a few words.  They are to make code more skimmable, not to describe every detail.
+- Code that is intuitively obvious when skimming (such as variable declarations or simple error checks at the start of a scope) does not need a section comment.
+- Avoid overusing section comments when they do not actually make code easier to skim.
+- For example:
```
nboard, line 577: **Section Comments**
Is this meant to be a positive example, or an example of overused section comments?

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+Any English sentence should use two spaces (or a newline/similar whitespace) between a period and the following sentence.  This makes it easy to visually jump between sentences and skim text.  This is not required for string literals, such as error messages in code, although these rarely have more than a single sentence anyway.
+
+### Warning Comments
+- Unintuitive warnings and "foot-guns" should be called out in brief, concise comments.
+- These should be very rare, only used if there isn't a better way to communicate the information.
+
+### Section Comments
+- Divide code into logical sections that start with brief comments for skimming code.
+- Sections should be separated by blank lines.
+- These comments should start with a capital letter and end with a period.
```
nboard: **Section Comments**
Given that these serve almost as section titles and are typically sentence fragments, I would advise against using periods

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+- Copyright notices are 74 characters wide, so a cursor on the right edge of one is in column 75.  This may need to vary slightly depending on the language's comment syntax.
+- Use spaces when spacing out a copyright notice, not tabs (see [indentation](#indentation)).
+
+This is the format in C:
+```c
+/************************************************************************/
+/* Centrallix Application Server System                                 */
+/* <Subsystem>                                                          */
+/*                                                                      */
+/* Copyright (C) <created>-<year> LightSys Technology Services, Inc.    */
```
nboard: **Copyright Notices**
Consider, "<year_created>-<year_modified>"

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+/* Description: <Module description>                                    */
+/*              <Can wrap to multiple lines>                            */
+/************************************************************************/
+```
+- Replace the values in angle brackets (`<...>`).
+- `<Subsystem>` names the part of the project the file belongs to, such as `Centrallix Core` for `centrallix/` or `Centrallix Base Library` for `centrallix-lib/`.  The line above it, `Centrallix Application Server System`, is the same in every file.
+- `<created>` is the year the file was created and `<year>` is the year it was last modified, so update the copyright notice when modifying a file with anything more than trivial changes.  When both are the same year, only write that year itself.
+
+### Markdown Files
+Markdown files use their own copyright notices which are always placed at the start of the file.  Interestingly enough, these are not comments because the entire markdown file is considered to be documentation so markdown comments are rarely needed.
```
nboard: **Markdown FIles**
The end of the sentence feels a bit tangential. Perhaps indicate that because md files are themselves documentation, hiding the copyright from the end user is unnecessary (assuming that is what you meant)

centrallix-sysdoc/B2CodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+**Author**: <author name>
+
+**Date**: <date created, e.g. June 24th, 2026>
+
+**License**: Copyright (C) <created>-<year> LightSys Technology Services.  See `LICENSE`.
+```
+- Replace the values in angle brackets (`<...>`).  For more info, see above.
+
+### XML & XSL Files
+XML files use a single-line notice rather than the full block because these files are also considered documentation, like [Markdown files](#markdown-files).
```
nboard: **XML & XSL Files**
This comment feels odd, since markdown files have more than a single line for their copyright notice

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -0,0 +1,650 @@
...
+
+## AI Agents
+For AI Agents reading this document for the first time, I recommend saving a memory saying where this doc is (so it can be used for reference), as well as brief notes on a couple frequently used styles like Indentation, Braces, Comments, Naming, and Error Handling so you don't have to reread it with every request.
+
+
+## Todo - Israel
+Styles that still need to be decided and documented:
+- How Markdown files are styled?
+- How Python files are styled.  The 26 files in `centrallix-ui-test/tests`.  They have no rules today, and several universal rules here do not fit them.
+- How should long expressions or multi-line conditions be broken up? Does the line end with or start with the operator, and how far are continuation lines are indented?
```
nboard, line 707: **TODO**
I would like to vote for starting lines with the `&&`/`||` operators. I feel it makes it immediately obvious when reading the line that it belongs to the expression above it, and makes it so deleting/adding lines effect fewer lines

centrallix-sysdoc/CodingStandards.md
nboard, line 1: The parts of this document that overlap with B2CodingStandards.md should probably be trimmed out (though, this can wait until the document is in a more final form to ensure you don't accidentally cut content from both documents)

nboard (commented):
This looks really promising overall! I may have gotten a bit too much into the weeds on word choice for this early in the project, but hopefully this is helpful overall.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: That's a good choice for generalization.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: That's a good point. However, I don't want to lose the commit history when I change the file, so I'll save that interactive rebase and force push for tomorrow.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: Agreed, but I forgot to make a todo for that. Added.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Done.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: > "which would solve the issue with not being able to use tabs to align rows in structs and the like"

I don't think it does. Wouldn't we need to _require_ a tab width for that? I also don't see the issue with requiring spaces for indenting structs.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: I thought having the space made code easier to read. There's a sizeable distinction between a function call and one of these language constructs, and the space makes that easier to see at-a-glance.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: Oops, I meant to require one pointer declaration per line for C and it looks like I forgot.

Lightning11wins: Technically, `char* a, b;` only declares 1 pointer, so you're following the rule, so I wrote it slightly stronger: "In C, a pointer cannot be declared on the same line as another variable".

The "In C" is important because technically all variables are pointers in a language like JS, and I don't see a good reason to prevent declaring multiple variables per line in JS. I doubt most people would think like that, but I can't confidently say that no LLMs would.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I went one step further and just replaced the makefile specific rule with this general one:

> Binary operators and assignments should have a space on either side of the operator.

I don't think we need to specify makefiles specifically since the reader *should* understand universal rules like this one to apply universally.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: I considered that, but the second sentence is coppied from Greg's original style and I thought it was fun, so I kept it in.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: True, but it's short and I thought I should be complete.

Also, this is a universal rule, but I can't think of any languages *with braces* where this isn't the case.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: Done, added an example for the section.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: I had Claude make it simpler because it turns out that I struggle to come up with good, simple examples. This new one seems good, though!

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: I just put to follow C as closely as the language allows (and JS allows this).  The consistency is helpful, and I don't see any reason to make devs switch styles for different language any more than is clearly beneficial.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I trimmed the explicit example.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Simplified.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Turns out prefixes were really inconsistently marked so I switched to a more consistent system for doing that in general.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Makes sense, although we still need to say why because this is a recomendation, not a requirement.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Done.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I believe so, that's why I put "(`.c` files)" on the function definitions line. Is there an issue with how I handled this?

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: You're right, this needs to be rewritten.

Lightning11wins: Done.

Lightning11wins: Redone. It's better now.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: I think that's obvious, although I would discorage it in C because it means you have to make an extra jump to find the parameters. Usually the struct is declaired in a .h file, so it's far away from the function definition. That feels like it slows down code reading.

For JS, on the other hand, this is a great pattern. I think every widget init function uses it.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Oops, forgot to write out my reasoning explicitly. I'll do that.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: It's because they share a line in the format. Adding explanation.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: This entire section needed a restyle, and the part about sizeof() belonged better in spacing. I've done that now and I think it reads easier.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Turns out there was a lot to explain and I did my best to keep it concise. This also doubles to warn people about what's up with ht_flex_format, which is a useful bonus benefit.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: > "not reuse the same variable for multiple purposes"

That's such bad practice that I forgot people were doing it. I thought it was implied by "Local variables should be declared as close to where they are used as possible", but it bears repeating, just to be sure.

Lightning11wins: > "Many functions have a char* ptr; declared at the start which is used to store various function results with very different purposes."

I do use that particular pattern sometimes because you often need a temporary buffer to receive strings that you might not want to store. That said, it's not a great practice so outlawing it isn't a bad idea.

Lightning11wins: > "that any variables that contain data that will need freed on error or at the end of the function need to be allocated at the start of the function"

I'm guessing you meant declared when you said allocated?

In that case, this a good thing to require, especially since C doesn't actually require this so you can get uninitialized reads and invalid frees pretty easily.

Lightning11wins: > "so we don't end up with code trying to determine if a variable that was potentially never even declared needs freed"

I have done this. Writing code such that you don't have to puts more restrictions but it leads to better code.

Lightning11wins: I've added two new rules which should fix all of this.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I thought you were advocating we outlaw the not operator for a second, but now it makes sense and I agree.

I've rewritten this with your wording.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Good point, moving it.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I'll move this in a separate commit.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I'll move this in a separate commit.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: > "declaration would have a comment explaining what the set was for"

This statement is a bit ambiguous because there's two declarations.

**Declaration 1**: The `typedef` that declairs the type for the constant set.
- I agree, this should have a comment explaining the set. I'll add that.

**Declaration 2**: Any variable declared with a value from the constant set.
- I don't think you should have to comment anything about this. Like any other variable declaration, it the variable name should make it obvious what the variable is for in most cases with no special comment needed.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Ohhhh, I completely forgot about error enums and error flags, which are an ideal usecause for negative ids! Fixing that.

Lightning11wins: Added a callout. This means it's technically legal to use negative types for flags, but I doubt anyone would do that.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: hmmm... This was the previous standard, I'm not exactly sure how to express or apply that new one because this is pretty theoretical.

It seems like the set should be able to explain its purpose in the module in most cases, anyway.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I mostly coppied this section from Greg's documentation because I don't understand it well and I guess that kind of shows.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Ah, so change it so that tags are never indented? I can do that.

Lightning11wins: This does make error handlers for non-function scopes look a bit strange, but I think that's fine.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: Mine are complete sentences and I don't like using Title Case for these.  I use the same style for commit messages (although they're usually a little longer because they sometimes breifly explain *why* I made a change).

Lightning11wins: Some of them are also a bit long to be considered "titles", like:

```c
/*** Compute the X offset at which the image is flush with
 *** parent's right content edge.
 ***/
```

These usually accompany code that's *maybe* too complex, but there's some logic that's just really hard to write that simply.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Fixed.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: I thought it was a good example and I only overused them a little bit because the place that these really shine is on code that's too complex/long to put in a good example. I rearranged the text to make this less confusing.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Done.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Done.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I wanted to put `const` first because it should be the more common and more desired choice.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Reworded. I'd like to keep the "scrolling a programmer must do" wording, though, because it emphasizes that this is work your forcing any good programmer to do to understand your code.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I reworded and I think the new wording is better.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Yeah, I really don't know how to do commentsfor xml files. I kind of had to make this up because there were no existing example, but I felt like they should get *something*. I'll just remove the "Like Markdown files".

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: I agree, but I think I remember cases where this caused issues.

Lightning11wins: Yeah, I struggled a lot with this code snippet:
```c
const bool font_changed = false
    || style->FontID != context->CurStyle.FontID
    || style->Color != context->CurStyle.Color
    || realComparePrecision(style->FontSize, context->CurStyle.FontSize, PRT_HTMLFM_FONTSIZE_PRECISION) != 0;
```
I don't really like the `false ||` / `true &&` style I end up using so that we don't have to indent one line differently from every other line.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Updated.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: I'll make a discussion for this one.

centrallix-sysdoc/B2CodingStyle.md
nboard: <del>That's fair, but really with the previous statement its a bit redundant, and I find the current sentence construction to be a bit vague... 
Maybe, "In `js`, use `const` as described above, and `let` for cases where the variable value can change."</del>
Scratch that, your change works fine

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: I generally preferred them without spaces, but I won't push on this one

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: Huh, interesting. I can't really ague with that then

centrallix-sysdoc/CodingStandards.md
Lightning11wins: Added to TODOs, I'll revisit this later. You can mark this as resolved.

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: Fair enough

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: My bad,  I forgot JS has those.

That said, I feel like maybe any language without a similar style of  multi-line comments (Python, SQL, structure,, etc.) should just ignore this section altogether? I feel like the natural thing in any other language would be to have the right, single, and multi line comments all the same. The later section on comments explains when/how to use comments content wise, which maintains some consistency at least

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: I forgot how much the changing tab alignments would throw it out of whack... We would have to continue to require a tab space width, although I would probably recommend 4 instead of the current 8. I take it you're pretty against that idea? 

Personally, I find spaces to be more work to upkeep. Any change earlier in the line always requires fixing the alignment, whereas tabs lower the chances of needing to readjust. I also find hitting tab 1-5 times easier than hitting space 4-10 times on every line (twice, if you have comments to align as well). 

The stability of using spaces has advantages, but I still lean towards tabs... What are your thoughts?

centrallix-sysdoc/B2CodingStyle.md
nboard: Adding more justification for optional features makes sense, so users can tell when to ignore or follow it, but I feel that, "consistency with other files, unless there is a good reason to do otherwise" is a bit too vague to be helpful, and applies to most rules in the document

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: Wow, that was actually *way* easier than I thought it would be. Just one amended commit during an interactive rebase, and it's renamed across the entire history.

Force push incomming.

centrallix-sysdoc/B2CodingStyle.md
nboard: I did not explain that well, my bad. 
Is 6 also meant to apply to prototypes found in .h files (where the function is not defined later in the same file), or is there not a recommended location for those?

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: Fixed.

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: > > "that any variables that contain data that will need freed on error or at the end of the function need to be allocated at the start of the function"
> 
> I'm guessing you meant declared when you said allocated?
> 
> In that case, this a good thing to require, especially since C doesn't actually require this so you can get uninitialized reads and invalid frees pretty easily.

Yes, declared, my bad

centrallix-sysdoc/B2CodingStyle.md
nboard: I meant the declaration of the list as a whole, which I believe fits **Declaration 1**. While ideally the flag or type names would make it clear what their individual purposes are, the reason for the set to exist at all often benefits from a comment. 

I also failed to make my point clearly... While I agree typecasting is a good idea to remove ambiguity, and using a typedef means you can reliably create variables to store the constants when needed, the bullet indicates the sole reason to do this is so that "declarations using this type will show what the value means", which I interpreted to mean it makes understanding the purpose of the set more obvious. 

Altogether, comments explain why the set exists, and there are better reasons to have a typedef (and subsequent casts) than clarification on the meaning, so I'm suggesting we could change the bullet to talk more about the advantages of the typedef/cast, or else remove that last part if you feel the justification isn't needed.

centrallix-sysdoc/B2CodingStyle.md
nboard: Interesting. It seems like there are some cases where a flag isn't related to a sturct at all (after a quick search, ht_render.h appears to have some at least), or seem to have uses beyond the struct its associated with, which is what made me think this rule didn't quire fit. Maybe add a note that flags independent of structs should just base the name off the module?

centrallix-sysdoc/B2CodingStyle.md
nboard: I think you did an excellent job overall, and it does make sense here, I just think it would be more helpful with the structs

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: Could we keep it sentence case, but just not requiring a period? It seems like most section comments in the codebase currently are fragments, which we could end with a period, but feels odd given that they would not be valid sentences regardless. 

The example you gave is actually a sentence fragment, which works perfectly well for its use and avoids unnecessary fluff. While periods make sense to separate multiple snetence fragments, ending with one feels mismatched with the rest of the technical-but-informal style to me.

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: Looks good!

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: Looks great and explains everything well

centrallix-sysdoc/B2CodingStyle.md
nboard: Great job on the rewrite!

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: That makes sense, and I think the example does work well

centrallix-sysdoc/B2CodingStyle.md
nboard: Looks good!

centrallix-sysdoc/B2CodingStyle.md
nboard: That works!

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: One issue with tabs is that you can't do all alignment with them and you need to keep falling back to spaces.  For example, the copyright comment starts every line with `/* `.  Your upkeep concern is a real downside, but it doesn't seem like a *huge* issue (basically no affect on git-blame history, just a few extra seconds wasted on alignment every time someone makes a change).  However, I don't think that's worth needing to mix tabs and spaces for alignment the way I think we'd need to do if we changed this policy.

As for requiring 4-space tabs so that it's possible to *have* aligned comments on indented code, I could agree with that.  I wanted to avoid adding requirements that didn't benefit us, but there are benefits to that one so I guess we could add it.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: By my reading of the requirements in this section, as soon as your language doesn't support `/*`/`/**`/`/***`, you're free to use `//` for everything.  I think that makes this issue already-resolved, right?  Or are you saying we should be more explicit about handling the exception cases that way?

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: Do you mean "Parameter found." because "found" is a past participle? I suppose you're technically right.

Lightning11wins: Not ending with period feels weird for the occasional longer section comments that have multiple sentences. For example:
```c
/*** Write HTML header.  Report content always sits on a white page area
 *** (white body, or the white page cell in paginated mode), so white is
 *** the current background.  This lets us skip setting the background on
 *** white cells, reducing the HTML size
 ***/
```
Long section comments like this one might be an abuse of the standard, but they feel like the most readable way to handle some complex situations.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Sounds good, although that makes this section even shorter. I think that's actually a good thing, though, because it should probably be merged with enum. These sets really feel like they're just a special type of enum.

Lightning11wins: Refactor done.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: That makes sense. I think the new version is better enough to resolve this.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Ohhhh, yeah, that could be improved. I changed it to "Function declarations".

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: I thought of a set of flags as its own type that may or may not be included in one or more struct definitions.

Lightning11wins: Wait, why is there a rule here that flags need to include the struct name?! Looks like I need to re-review this section.

Lightning11wins: Updated.

centrallix-sysdoc/B2CodingStyle.md
Lightning11wins: Ahh, I agree. That was badly worded.

I reviewed the wording in this entire section and updated this part to read:

> The set type should have a `typedef` specifying a numerical type.  This makes declarations using this type more explicit and obvious.

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: I guess that final comment about the asterisk tiers not applying to XML specifically seems to imply that other languages are able to use their own "comment syntax its language accepts" (so the code compiles/runs/is parse-able), but are still expected to add extraneous `*` to comments to match the tiers. Is that the intention? So say, python would do:
right: `#* comments here *`
Section: end of line: `#** comments here **`
Multi line: 
```python
#*** Comments here
#*** more comments
#*** more still
#***
```
That actually looks better than I expected, but I feel like just adding more pound signs would work better (and could also fit a reading of the current rules), or just stick to a single pound sign most of the codebase uses

Lightning11wins: Yeah, I need to make that exception more explicit. On it.

Lightning11wins: Done and pushed.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: Added:

> It is highly recommended to set your editor tab length to 4 spaces.  When writing code, you may assume that all readers will view it using 4-space tabs.

centrallix-sysdoc/CentrallixCodingStyle.md
nboard: I don't think we need to forbid periods, I just think requiring them is counterproductive

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -30,8 +41,8 @@ These rules apply to every language in Centrallix, unless a section states other
 
 ### Indentation
 - Each level of indentation should use a single tab.  Do not use spaces for general indentation.
-- You can set your editor tab length to any size you prefer.  However, code should look reasonable with any tab length up to 8 spaces.
-- Prefer tabs, but always use spaces for aligning characters, such as in a copyright notice or a column of struct members.  Tabs break alignment when the viewer uses a different tab length.
+- It is highly recommended to set your editor tab length to 4 spaces.  When writing code, you may assume that all readers will view it using 4-space tabs.
+- Prefer tabs, but always use spaces for aligning characters, such as in a copyright notice or a column of struct members.
```
nboard: **Indentation**
This sentence is a little confusing to me, maybe "Always use spaces when aligning columns to the right of text, such as in..."
The first bullet already requires tabs for general indentation, but you can always add a second sentence or separate bullet to emphasize the use of tabs if desired.

centrallix-sysdoc/CentrallixCodingStyle.md
```diff
@@ -203,6 +221,7 @@ For: `.c`, `.h`
 - Each member is declared on its own line.
 - Member names are aligned in a column, padded with spaces (never tabs, see [indentation](#indentation)).
 	- Leave a few extra spaces of padding where a longer type may be added later, especially in a struct whose types are all short or which has many members.  This allows adding the type later without a reflow on every line that buries the real change in git-blame.
+- If a structure can represent kinds or types (not data types, this is a conceptual thing), it almost always holds a field using a [set type](#set-types) (typically an [enum](#enums)) to store data has been stored.
```
nboard: **Include Files:** line 224
I don't understand what the end of this line is saying. Did you mean, "to store the type/kind of the stored data"?

centrallix-sysdoc/CodingStandards.md
Lightning11wins: I'll revisit this after Greg's review ensures that the standard won't change significantly in the near future.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: I think we should have a standard for consistency because alternating randomly is probably worse than either choice. Should we just recommend using periods and leave the decision as a matter of taste?

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: You're absolutely right—"prefer tabs" buys us nothing here. The rest of the sentence is load-bearing—we can remove that wording.

Lightning11wins: Fixed.

centrallix-sysdoc/CentrallixCodingStyle.md
Lightning11wins: Yeah, that's what I meant. This line was a real struggle to write and I rewrote it a lot. Fixed.

Lightning11wins:
The next step is to run a linter with these new rules. However, that merge-conflict all open PRs. As a result, **this PR is now soft-blocked from continued development by *all open Centrallix PRs***.

Lightning11wins:
**TODO:** Reconsider naming conventions for internal functions. I don't really like `mss_i_error()`.

